PR #7, opened 2021-05-15
refactor: remove state from structs that differentiate behaviors

ereslibre:


src/registry/mod.rs
```diff
@@ -61,11 +42,95 @@ impl Registry {
...
+            }
+            Err(err) => Err(anyhow!("could not download Wasm module: {}", err)),
+        }
+    }
+
+    async fn do_push(
+        mut client: Client,
+        policy: &[u8],
+        reference: &Reference,
+        _registry_auth: &RegistryAuth,
```
flavio, line 111: Why aren't we using the `RegistryAuth`? The code seems to be hard coded to perform only anonymous push operations. Am I missing something?

flavio (changes requested):
I just have some questions around push operations performed against registries that require authenticated users.

flavio (approved):
LGTM

PR #8, opened 2021-06-01
Fix ghcr push

flavio:
This PR fixes the push operations against ghcr and against generic registries secured by TLS.

The oci-distribution crate was affected by a bug that caused pushes against ghcr.io (and probably other registries) to fail. Our code now builds against a fork of oci-distribution. This will be changed once upstream publishes a new version of the crate that includes this fix.

Finally, prior to this commit, the `push` code was nevertheless broken when dealing with registries secured by TLS. The code would push over TLS and then, even when the push completed successfully, would perform another push over plain HTTP. The second push is not needed, plus it always leads to a failure (pushing over plain HTTP against a HTTPS endpoint).

Once this PR is merged I'll tag a new release of policy-fetcher

### DIFF
--- a/src/registry/mod.rs
+++ b/src/registry/mod.rs
@@ -55,13 +55,21 @@ impl Registry {
             Reference::from_str(url.as_ref().strip_prefix("registry://").unwrap_or_default())?;
         let registry_auth = Registry::auth(reference.registry(), docker_config);
 
+        // First we try to push to the registry using TLS
         if let Err(err) = Registry::push_tls(policy, &reference, &registry_auth).await {
             let host_and_port = crate::host_and_port(&url)?;
             if sources.map(|sources| sources.is_insecure_source(host_and_port)) != Some(true) {
+                // Push failed, plus the registry is not marked as "insecure" -> time to bubble up
+                // the error
                 return Err(anyhow!("could not push Wasm module: {}", err));
             }
+        } else {
+            return Ok(());
         }
 
+        // We are here because pushing to the registry using TLS didn't work,
+        // but the registry is marked as insecure. We will do one last attempt
+        // and push over plain HTTP
         Registry::push_plain(policy, &reference, &registry_auth).await
     }
 
--- a/src/registry/mod.rs
+++ b/src/registry/mod.rs
@@ -9,7 +9,6 @@ use oci_distribution::{
     Reference,
 };
 use std::{path::Path, str::FromStr};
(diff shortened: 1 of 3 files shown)


src/registry/mod.rs
```diff
@@ -114,7 +123,6 @@ impl Registry {
                 registry_auth,
                 None,
             )
-            .compat()
```
ereslibre, line 117: Are you sure this doesn't explode in runtime? I remember an issue with futures in one version, and the future executor in another one, that was why we needed the `compat()` call.

flavio:
This is the upstream PR that fixes the issue inside of oci-distribution: https://github.com/deislabs/krustlet/pull/617

src/registry/mod.rs
```diff
@@ -25,6 +25,7 @@ impl Registry {
     fn client(client_protocol: ClientProtocol) -> Client {
         Client::new(ClientConfig {
             protocol: client_protocol,
+            ..Default::default()
```
ereslibre, line 28: `ClientConfig` only has `protocol`

flavio: Not with the last round of patches that got merged into oci-distribution :)

flavio: Remember, we're now building against the `main` branch of oci-distribution. The documentation on crates.io is outdated.

ereslibre: Right!

src/registry/mod.rs
ereslibre: `oci-distribution` was using returning an older Tokio future that was not executable by the runtime in our `policy-server` back in time: https://github.com/kubewarden/policy-fetcher/blob/b73d3e09c5ced9157960ca1c3806723b7774a3ab/src/registry/mod.rs#L134-L136

I think it would be good to revisit the current situation. I was expecting this to fail if no `compat()` was called here either.

flavio: I just removed the `tokio-compat-02` crate from `Cargo.toml`, plus all the `compat()` occurrences. Everything is building fine

ereslibre: The problem is not with building, but with panic runtimes. If everything is good now, then fine!

flavio: Since Feb 2021, oci-distribution is built against tokio 1.0, which makes the tokio-compat code useless. I've updated the PR to remove this dependency

ereslibre (approved):
Thank you!

PR #10, opened 2021-06-14
feat: ignore invalid host configuration on the dockerconfigjson

ereslibre:
Fixes: https://github.com/kubewarden/kwctl/issues/47

`kubectl` has generated docker config json files that miss the `auth`
key in the `auths` field [1].

Due to this, make the parsing of the dockerconfigjson file less
strict, with an optional `auth` field. If this field is not present,
the host will be ignored, but the rest of the file will apply.

We keep the same behavior of a complete failure in the following
cases:

- The JSON file is not syntactically valid
- The basic auth information is not valid base64
- The decoded base64 is not in the form `username:password`.

[1] https://github.com/kubernetes/kubectl/issues/571

### DIFF
--- a/src/registry/config.rs
+++ b/src/registry/config.rs
@@ -17,14 +17,14 @@ pub struct DockerConfigRaw {
     auths: HashMap<String, RegistryAuthRaw>,
 }
 
-#[derive(Clone, Debug)]
-pub(crate) enum RegistryAuth {
+#[derive(Clone, Debug, PartialEq)]
+pub enum RegistryAuth {
     BasicAuth(Vec<u8>, Vec<u8>),
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq)]
 pub struct DockerConfig {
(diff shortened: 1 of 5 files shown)


flavio (changes requested):
Everything looks good. One minor thing: can you extend the unit tests to have also a test case parsing a file with the expected auth entry?

ereslibre:
@flavio: updated for adding some integration tests in `tests` (as recommended by the book in https://doc.rust-lang.org/book/ch11-03-test-organization.html#integration-tests), with the `test_data` folder containing the files to be unmarshaled for the tests.

I think it's fine, but one shortcoming I see is that the types used in the crates within `tests` (every test is an individual crate created by `cargo test`) need to be `pub`, as `pub(crate)` is not enough. Maybe something to take into account in the future, especially if we write crates that can be consumed by third parties. I think in this case is fine, because this crate was always meant to be internal to `policy-server` and `kwctl`.

flavio (approved):
Looks great, thanks for all the new changes

PR #12, opened 2021-06-17
feat: finish implementation of insecure sources

ereslibre:
The policy fetcher implements the retry logic in case of an insecure
registry:

- Try with defaults + custom CA if any
- If the previous step failed and the registry is insecure, try with
  TLS, but ignoring certificate check
- If the previous step failed and the registry is insecure, try with
  plain HTTP


src/registry/mod.rs
```diff
@@ -82,22 +124,25 @@ impl Registry {
...
+        };
 
         match image_content {
             Ok(image_content) => {
                 let mut file = File::create(destination).await?;
                 file.write_all(&image_content[..]).await?;
                 Ok(())
             }
-            Err(err) => Err(anyhow!("could not download Wasm module: {}", err)),
+            Err(err) => Err(anyhow!("could not pull policy: {}; make sure the registry is trusted or in the insecure list and that the policy is present", err)),
```
flavio, line 145: Do we need all these explanations? Isn't the proper failure message already stored into `err`?

src/registry/mod.rs
```diff
@@ -143,34 +189,113 @@ impl Registry {
...
+            Some(sources) => {
+                if let Some(custom_certificate_authority_chain) =
+                    sources.source_authority(reference.registry())
+                {
+                    custom_certificate_authority_chain
+                        .iter()
+                        .map(|certificate| Certificate {
+                            encoding: CertificateEncoding::Pem,
+                            data: certificate.0.clone(),
+                        })
```
flavio, line 209: Isn't there any Err raised when creating a certificate? Moreover I see we have `Pem` hard coded. Can we try to load with `Pem`, if it fails try with `Der` and if that fails too bubble up the error?

flavio (approved):
Overall looks good to me. I just left some questions

src/registry/mod.rs
ereslibre: Yes, the problem I see with bubbling all errors in a "blind" way is that the user might read "invalid OpenSSL connection", but they are lacking a lot of context of what really happened. By wrapping the error we can provide extra context local to us, so we know where the error comes from without the need of a stacktrace, and also give the user more information that can help them to look into the right places.

ereslibre: Or we could add a less detailed error message here and complete on higher places closer to the user. Is this what you meant?

flavio: I mean that `image_content` here is coming from the lines 127+ that you just added.

An error caused by pulling from an insecure TLS endpoint is already bubbled up to the user via the `return Err(..."could not pull policy: {}; not retrying in unsafe mode because registry {} is not insecure").

The only error we get is the one from `Registry::fetch_plain`

Aren't we adding more "noise" here? I would just drop "make sure the registry is trusted or in the insecure list and that the policy is present"

ereslibre: Ugh, brain fart, you are right.

src/registry/mod.rs
ereslibre: This is just the initialization of the object, but let me check if we can do something like that you describe. I think it would be good to support both.

flavio: That, plus ensure we tell the user we can't parse one of the certificates he provided us

ereslibre: I don't think we should do that. They give the certificate in the sources file, we try with pem, then with pem. If both fail our error is: this certificate is neither a PEM nor a DER

flavio (approved):
Wow, this is super neat! :clap:

PR #14, opened 2021-07-06
Pull from registry by default

flavio:
When the user doesn't specify a url scheme, use the `registry://` one.

This makes possible to copy and paste policy urls from hub.kubewarden.io into `kwctl pull`

This provides a smoother UX


ereslibre (approved):
:clap:

PR #16, opened 2021-10-04
Check host+port when checking for insecure hosts

ereslibre:
If port is present, check `host+port` in the insecure hosts mapping to
determine if we should retry the pull/push.


viccuad (approved):
nice! :)

PR #17, opened 2021-10-19
Allow certificates to be put inside of `sources.yml`

flavio:
Change how source authorities are expressed
    
Change the `sources.yaml` file to allow users to embed the whole certificates inside of it. This is a change required to simplify the kubewarden-controller.
    
This change breaks the previous syntax used by `sources.yml`:
* It's now possible to read certificates from file, like before, but the yaml objects are expressed differently
* The previous code allowed multiple certificates to be associated with a FQDN, but this was never documented. After this change each domain has just one certificate associated with it, which is also what our documentation showed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "policy-fetcher"
-version = "0.1.14"
+version = "0.1.15"
 authors = [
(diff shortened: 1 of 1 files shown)


flavio:
Done, I've pushed a tag for this new version of the crate

PR #25, opened 2021-11-12
feat!: Add `verifier::verify_local_file_checksum()`, return manifests on `verifier::verify()`

viccuad:
Add `verifier::verify_local_file_checksum()`. As input, it uses a verified manifest digest. That way the callers are
aware that this function should only be called on verified policies.

Refactor with breaking change: Return verified manifest digest in `verifier::verify()`, and don't silently compare checksums there.

Needed by https://github.com/kubewarden/kwctl/pull/84.

### DIFF
--- a/src/verify.rs
+++ b/src/verify.rs
@@ -36,10 +36,9 @@ impl Verifier {
     /// When annotations are provided, they are enforced with the values specified
     /// inside of the Sigstore signature object.
     ///
-    /// Finally, this method checks whether the specified policy already exists
-    /// on the local file system. When that happens, the checksum of the local file
-    /// is compared with the one mentioned inside of the signed (and verified) manifest.
-    /// That ensures nobody tampered the local policy.
+    /// Note well: this method doesn't compare the checksum of a possible local
+    /// file with the one inside of the signed (and verified) manifest, as that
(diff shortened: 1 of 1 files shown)


src/verify.rs
```diff
@@ -31,25 +31,25 @@ impl Verifier {
...
     }
 
-    /// Verifies the given policy using the verification key provided by the user.
+    /// Verifies the given policy using the verification key provided by the
+    /// user.
     ///
-    /// When annotations are provided, they are enforced with the values specified
-    /// inside of the Sigstore signature object.
+    /// When annotations are provided, they are enforced with the values
+    /// specified inside of the Sigstore signature object.
```
flavio, line 38: Add also a comment explaining what is the `String` that is returned by the method

flavio (approved):
Everything looks great, I just left a minor suggestion

src/verify.rs
viccuad: Whoops, indeed! Thanks for the review, done in https://github.com/kubewarden/policy-fetcher/pull/25/commits/082b795196356088a8b14c0b80128f5a728c2273.

flavio (approved):

PR #26, opened 2021-11-16
expose fetch manifest digest

flavio:
Expose a method to fetch the manifest digest. This is needed so that we can expose this function to our policies.

While at that, I also did some refactoring of function params: from `&Option<T>` to `Option<&T>`.

### DIFF
--- a/src/registry/mod.rs
+++ b/src/registry/mod.rs
@@ -84,7 +84,7 @@ impl From<ClientProtocol> for ClientConfig {
 impl Registry {
     pub fn new(docker_config: Option<&DockerConfig>) -> Registry {
         Registry {
-            docker_config: docker_config.map(|dc| dc.clone()),
+            docker_config: docker_config.cloned(),
         }
     }
 
@@ -119,7 +119,7 @@ impl Registry {
             Reference::from_str(url.as_ref().strip_prefix("registry://").unwrap_or_default())?;
 
         let registry_auth = Registry::auth(reference.registry(), self.docker_config.as_ref());
-        let sources: Sources = sources.map(|s| s.clone()).unwrap_or_default();
+        let sources: Sources = sources.cloned().unwrap_or_default();
         let cp = crate::client_protocol(&url, &sources)?;
 
         let (m, _) = Registry::client(cp)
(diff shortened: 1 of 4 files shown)


flavio:
This is now ready for review.

src/registry/mod.rs
```diff
@@ -109,17 +109,18 @@ impl Registry {
...
-        sources: &Option<Sources>,
+        sources: Option<&Sources>,
     ) -> Result<oci_distribution::manifest::OciManifest> {
         let url = Url::parse(url).map_err(|_| anyhow!("invalid URL: {}", url))?;
         let reference =
             Reference::from_str(url.as_ref().strip_prefix("registry://").unwrap_or_default())?;
 
         let registry_auth = Registry::auth(reference.registry(), self.docker_config.as_ref());
-        let cp = crate::client_protocol(&url, &sources.clone().unwrap_or_default())?;
+        let sources: Sources = sources.cloned().unwrap_or_default();
```
ereslibre, line 122: Given `manifest` returns a `Result`, I'd bubble the error up (or map the error) if `cloned()` fails instead of silently defaulting the structure.

src/registry/mod.rs
```diff
@@ -128,24 +129,33 @@ impl Registry {
         Ok(m)
     }
 
-    pub async fn push(&self, policy: &[u8], url: &str, sources: &Option<Sources>) -> Result<()> {
+    /// Fetch the manifest's digest of the OCI object referenced by the given url.
+    /// The url is expected to be in the "registry://" format.
+    pub async fn manifest_digest(&self, url: &str, sources: Option<&Sources>) -> Result<String> {
+        let url = Url::parse(url).map_err(|_| anyhow!("invalid URL: {}", url))?;
+        let reference =
+            Reference::from_str(url.as_ref().strip_prefix("registry://").unwrap_or_default())?;
```
ereslibre, line 137: I think we want to bubble the error up as well here, falling back to the default would end up in `Reference::from_str(String::default))` what could lead to errors hard to understand on the user side.

src/registry/mod.rs
```diff
@@ -128,24 +129,33 @@ impl Registry {
...
+
+        let registry_auth = Registry::auth(reference.registry(), self.docker_config.as_ref());
+        let sources: Sources = sources.cloned().unwrap_or_default();
```
ereslibre, line 140: We should bubble up here as well I think.

ereslibre (changes requested):
LGTM except for some comments.

viccuad (commented):
LGTM modulo the points raised by Rafa, which I agree with.

src/registry/mod.rs
ereslibre: NVM, I just realized `cloned()` just returns an `Option` on `Option`. I see, is a way to not consume the `Option` because `unwrap_or_default()` consumes it, right?

In that case, discard the rest of my comments in the same direction.

flavio: yes, that's the case. I'll resolve the other comments you left

src/registry/mod.rs
flavio: Good point, I have to admit I copied and pasted this snippet from other parts of the code. I've changed it now, can you take another look?

ereslibre (approved):

src/registry/mod.rs
viccuad: doh, indeed, now I see it too. Thanks both!

viccuad (approved):

PR #27, opened 2021-11-25
Handle operations against Docker Hub

flavio:
Some of the methods of the `Registry` struct are going to be used against container images, not kubewarden policies.

For example, this will happen to the methods dealing with OCI objects manifests (regardless of them being images or artifacts).
A policy might rely on this method to obtain the digest of a container image that is mentioned inside of a Pod `spec` section.

We have no control over these container images, they could also be images coming from the Docker Hub or even from the "Docker Hub library".
That means that we must be able to handle OCI object names like `busybox`, `opensuse/leap`.

This commits makes `policy-fetcher` rely on a forked version of `oci-distribution`.
This is needed because some important fixes related with image parsing have not yet been released.

Some parts of our code had to be changed. We should not build a `Url` starting from the user input, but rather use `oci-distribution` to resolve the image name and then feed that into a `Url`.

Why that? A simple example: prior to this commit, a user input of `registry://busybox:latest` would have caused an error when building the `Url` object. That's because this string is not a properly formatted
url.

We have instead to:

  1. Resolve `busybox:latest` using `oci-distribution`.
  2. This will return `docker.io/busybox/library:latest` as image name.
  2. Create a `Url` with the fully resolved name.

Note well, the new image resolution algorithm of `oci-distribution` automatically appends the `:latest` tag to an image when needed.
Because of that some of our code had to be removed/refactored.

This is needed as part of the epic that plans to expose "image manifest related" operations to our policies.

### DIFF
--- a/src/registry/mod.rs
+++ b/src/registry/mod.rs
@@ -11,7 +11,9 @@ use oci_distribution::{
     secrets::RegistryAuth,
     Reference,
 };
-use std::{convert::From, path::Path, str::FromStr};
+use std::convert::TryFrom;
+use std::{path::Path, str::FromStr};
+use tracing::debug;
 use url::Url;
 
 use crate::fetcher::{ClientProtocol, PolicyFetcher, TlsVerificationMode};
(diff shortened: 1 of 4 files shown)


ereslibre (approved):

src/registry/mod.rs
```diff
@@ -246,3 +257,82 @@ impl PolicyFetcher for Registry {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use rstest::rstest;
+
+    #[rstest(
```
ereslibre, line 266: Nice 📚

PR #28, opened 2021-12-02
Replace openssl with rustls

ereslibre:


ereslibre:
Cannot request a review from @kubewarden/kubewarden-developers proper...

jvanz (approved):
LGTM! :D

src/verify.rs
```diff
@@ -21,14 +21,17 @@ pub struct Verifier {
...
     /// later used to interact with remote OCI registries.
-    pub fn new(sources: Option<Sources>) -> Self {
+    pub fn new(sources: Option<Sources>) -> Result<Self> {
         let client_config: sigstore::registry::ClientConfig =
             sources.clone().unwrap_or_default().into();
-        let cosign_client = sigstore::cosign::Client::new(client_config);
-        Verifier {
+        let cosign_client = sigstore::cosign::ClientBuilder::default()
+            .with_client_config(client_config)
+            .build()
```
flavio, line 29: I think this is not going to be enough. You have to provide a location for the rekor public key and for fulcio's certificate.

This is something the sigstore-rs library requires on the main branch (which is what you're building from, isn't it?).

You can hard code `~/.sigstore/root/targets/rekor.pub` and `~/.sigstore/root/targets/fulcio.crt.pem`. These files can be generated by `cosign initialize`

flavio (commented):
Overall good, I left one comment

src/verify.rs
ereslibre: Done. Please have a look. I think as a first approach this is fine. I'm hardcoding to `~/.sigstore/root/targets/{fulcio.crt.pem,rekor.pub}`, but if the home directory could not be inferred, then we default to the current directory.

Despite I know nothing about sigstore, I think we will want to make this parameter configurable on policy-fetcher, right? So policy-fetcher consumers can provide (if they wish) specific paths to where to read this certificate and key from, and if they are not found, either default to known locations or error out.

I can take care of adapting later if desired.

flavio: > Despite I know nothing about sigstore, I think we will want to make this parameter configurable on policy-fetcher, right? So policy-fetcher consumers can provide (if they wish) specific paths to where to read this certificate and key from, and if they are not found, either default to known locations or error out.

Yes, this is something we have to do, regardless of sigstore-rs default behaviour. Do you want to do that right now or later?

Also, I think we should handle better the case when none of these files exists. We should print a better error message and suggest to install `cosign` and run `cosign initialize`

ereslibre: > Yes, this is something we have to do, regardless of sigstore-rs default behaviour. Do you want to do that right now or later?

I can adapt this PR.

> Also, I think we should handle better the case when none of these files exists. We should print a better error message and suggest to install cosign and run cosign initialize

I think this message belongs to a tool like kwctl, but I don't think it belongs to policy-server -- the usual way in which is deployed is not as "easy" as running cosign initialize in the policy-server side. Therefore, I think this message does not belong to the library itself.

What I plan to do is to shape the error type of the Result, so we can specify the error there. Then, the crate consumer can write to their users the message that fits best for them. In `kwctl` it would be something along the lines you have proposed, on `policy-server` something without suggesting an action, just the error and why it's erroring.

WDYT?

flavio: I think the end user of the library (kwctl/policy-server) have to pass the files down to the policy-fetcher library. The validation of existence of the files should be done inside of the kwctl/policy-server, where the right error message is going to be created.

ereslibre: Updated.

src/verify.rs
```diff
@@ -21,14 +19,23 @@ pub struct Verifier {
 impl Verifier {
     /// Creates a new verifier using the `Sources` provided. These are
     /// later used to interact with remote OCI registries.
-    pub fn new(sources: Option<Sources>) -> Self {
+    pub fn new(
+        sources: Option<Sources>,
+        fulcio_cert: Vec<u8>,
+        rekor_public_key: String,
```
flavio, line 25: nitpicking: we can use `&[u8]` and `&str`

ereslibre: I know. I always have this doubt of who owns what. For example, sigstore-rs takes that, but does a `to_owned` that will result in cloning usually.

The question from the API perspsective is: if it's a parameter that will be used only once, the user can instantiate it, and pass ownership to the library. No need to clone. Passing a reference to immediately clone could seem strange, why not let the user decide and clone if they need, or pass the ownership, so they can avoid cloning?

In any case, I'm just forwarding the ref here, but I just wanted to note that is a thought decision, and that there are some nuances when it comes to deciding what type of parameters are wanted depending on the use case.

flavio (approved):
LGTM

PR #30, opened 2022-01-11
Change verifier API

flavio:
The verification of the signed manifest digest and the digest of the local `.wasm` file cannot be done using the Policy URL (e.g.: `registry://ghcr.io/kubewarden/foo:latest`).

While `kwctl` downloads all the policies into the central store (making possible to find the path to the wasm file, starting from the URI), the Policy Server can download policies into arbitrary locations.

The API of the verifier has been changed to take a `Policy` object, instead of a `str` containing the URI.
The Policy object has already two attributes: the URI and the path to the `.wasm` file.

This change is required to implement Sigstore verification inside of Policy Server.

### DIFF
--- a/src/verify.rs
+++ b/src/verify.rs
@@ -1,5 +1,4 @@
 use crate::sources::Sources;
-use crate::store::Store;
 use crate::{policy::Policy, registry::config::DockerConfig};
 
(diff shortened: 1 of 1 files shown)


viccuad (approved):
LGTM

PR #31, opened 2022-01-19
Update oci-distribution and sigstore-rs deps

flavio:
Use the latest releases of these crates.

Bump policy-fetcher to v0.3.1


viccuad (approved):
LGTM, thanks!

PR #32, opened 2022-01-20
push return immutable ref

flavio:
Change the API of `Registry::push` to return a value inside of its `Result` response.

The method now returns the immutable reference of the policy that has been pushed.

This information is important, because it allows kwctl users to use cosign (from Sigstore) to sign the **actual** policy that has just been pushed to the repo.

Signing policies by tag is not secure, there could be race conditions or attacks that could lead the user to sign a different content. Signing by immutable ref is not affected by these issues.

Because of this API change, the version of this crate will deserve a minor version bump: `0.3.x` -> `0.4.0`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,9 @@ async-trait = "0.1.50"
 base64 = "0.13.0"
 directories = "3.0.2"
 lazy_static = "1.4.0"
-oci-distribution = { version = "0.8.1", default-features = false, features =  ["rustls-tls"] }
+#oci-distribution = { version = "0.8.1", default-features = false, features =  ["rustls-tls"] }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -17,7 +17,9 @@ async-trait = "0.1.50"
 base64 = "0.13.0"
 directories = "3.0.2"
 lazy_static = "1.4.0"
-oci-distribution = { version = "0.8.1", default-features = false, features =  ["rustls-tls"] }
+#oci-distribution = { version = "0.8.1", default-features = false, features =  ["rustls-tls"] }
```
viccuad, line 20: leftover comment (yaay I'm not the only one)

flavio: Fixing and merging :)

PR #40, opened 2022-01-26
Windows path fixes

ereslibre:
Fixes: https://github.com/kubewarden/kwctl/issues/105
Fixes: https://github.com/kubewarden/kwctl/issues/109

Fix policy store path handling in Windows:

- Replace colons (illegal character name for paths) with a fake colon. This allows us to not do anything special for Windows in terms of policy store structure.
- Always use forward slashes for URL's

### DIFF
--- a/src/store.rs
+++ b/src/store.rs
@@ -1,6 +1,10 @@
-use anyhow::Result;
+use anyhow::{anyhow, Result};
+#[cfg(target_os = "windows")]
+use base64;
 use directories::ProjectDirs;
 use lazy_static::lazy_static;
+use path_slash::PathBufExt;
+use std::ffi::{OsStr, OsString};
 use std::path::{Path, PathBuf};
 use url::Url;
 use walkdir::WalkDir;
@@ -68,13 +72,27 @@ impl Store {
         }
     }
 
-    pub fn ensure(&self, path: &Path) -> Result<()> {
-        std::fs::create_dir_all(&self.root.join(path)).map_err(|e| e.into())
+    // Creates all directories provided in `path` starting from the
+    // root of this store.
+    pub fn ensure(&self, path: &Path) -> std::io::Result<()> {
+        std::fs::create_dir_all(&self.root.join(path))
     }
 
+    // Returns the full path of a policy coming from the URL `url` in
+    // this store. If `policy_path` is set to `PrefixOnly`, the
+    // filename of the policy will be omitted, otherwise it will be
+    // included.
     pub fn policy_full_path(&self, url: &str, policy_path: PolicyPath) -> Result<PathBuf> {
         let url = Url::parse(url)?;
         let filename = url.path().split('/').last().unwrap();
+
+        // In Windows we encode the filename with base64, so it can
+        // contain special characters like `:` that are not allowed in
+        // the filesystem
+        #[cfg(target_os = "windows")]
+        let filename: String =
+            base64::encode_config(filename.to_string().as_bytes(), base64::URL_SAFE_NO_PAD);
+
         let policy_prefix = self.policy_prefix(&url);
         Ok(match policy_path {
             PolicyPath::PrefixOnly => self.root.join(policy_prefix),
@@ -82,7 +100,11 @@ impl Store {
         })
     }
 
-    pub(crate) fn policy_prefix(&self, url: &Url) -> PathBuf {
+    // Returns the host and port (if any) as a string.
+    //
+    // On Windows, given that filesystem directories and files cannot
+    // include a colon (`:`), base64 the result.
+    fn host_and_port(url: &Url) -> String {
         let host_and_port = url
             .host_str()
             .map(|host| {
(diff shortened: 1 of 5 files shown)


viccuad (commented):
The code looks fine, but I really wonder about using an utf8 char that looks like a colon.
If I were a Windows user, and I need to type the path, it wouldn't be obvious from the start, and unless I copy-paste the filename after a `dir` it may be painful.

I wonder how others have worked around this. Maybe an escape char? or a different one altogether?

ereslibre:
> I wonder how others have worked around this. Maybe an escape char? or a different one altogether?

It depends. Docker for example follows a complete different storage scheme where they use the SHA's of the layers to store this information, and then the name of repos/images and tags is just data on top of that. We, for the sake of simplicity went with the 1:1 filesystem/URL hierarchy mapping.

I have been discussing with @nunix because another option would be to have a different layout on Windows, such as:

```
/home/ereslibre/.cache/kubewarden/store
└── registry
    └── ghcr.io
        └── kubewarden
            └── policies
                └── pod-privileged
                    └── v0.1.9
```

as opposed to the one in Unix:

```
/home/ereslibre/.cache/kubewarden/store
└── registry
    └── ghcr.io
        └── kubewarden
            └── policies
                └── pod-privileged:v0.1.9
```

I think this might be the most clear solution but it involves a bit of more special case handling for the Windows case.

We (@nunix and I) have settled on keeping this PR and aim for the hierarchical change in the near future.

ereslibre:
This implementation has been updated as discussed on yesterdays standup.

On Unix systems, the store is unchanged. On Windows systems, base62 is used for the name of directories and folders.

src/store.rs
```diff
@@ -68,21 +73,25 @@ impl Store {
...
+    pub fn ensure(&self, path: &Path) -> std::io::Result<()> {
+        std::fs::create_dir_all(&self.root.join(path))
     }
 
     pub fn policy_full_path(&self, url: &str, policy_path: PolicyPath) -> Result<PathBuf> {
         let url = Url::parse(url)?;
         let filename = url.path().split('/').last().unwrap();
+
+        #[cfg(target_os = "windows")]
+        let filename: String = base_62::encode(filename.to_string().as_bytes());
```
flavio, line 85: Add some documentation to the function that explains the purpose of this function but, most important of all, the different behaviour on windows. What are we doing on Win and why.

src/store.rs
```diff
@@ -93,49 +102,94 @@ impl Store {
                 }
             })
             .unwrap_or_default();
+
+        #[cfg(target_os = "windows")]
+        let host_and_port = base_62::encode(host_and_port.to_string().as_bytes());
```
flavio, line 107: Add some documentation to the function that explains the purpose of this function but, most important of all, the different behaviour on windows. What are we doing on Win and why.

src/store.rs
```diff
@@ -93,49 +102,94 @@ impl Store {
...
         let elements = url.path().split('/');
         let path = elements
             .skip(1)
             .take(element_count - 2)
             .collect::<Vec<&str>>()
             .join("/");
-        Path::new(url.scheme()).join(&host_and_port).join(&path)
+
+        #[cfg(target_os = "windows")]
+        let path: String = base_62::encode(path.as_bytes());
```
flavio, line 122: Add some documentation to the function that explains the purpose of this function but, most important of all, the different behaviour on windows. What are we doing on Win and why.

src/store.rs
```diff
@@ -93,49 +102,94 @@ impl Store {
...
+                                        {
+                                            Ok(acc.join(decoded_filename))
+                                        } else {
+                                            Err(anyhow!("invalid filename"))
+                                        }
+                                    })
+                                    .map_err(|_| anyhow!("cannot compute policy path"))?;
+                                #[cfg(any(target_os = "linux", target_os = "macos"))]
+                                let policy_store_path =
+                                    PathBuf::from("/").join(policy_store_path.collect::<PathBuf>());
```
flavio, line 175: I think it would be better to have this logic moved into dedicated functions. Both functions would have the same name, but only one would be visible depending on the `target_os`

flavio (changes requested):
Overall good. I left some comments.

Can you also add some unit tests? It would be great to have some way to test the encoding produced by these "build the path" functions.

src/store.rs
ereslibre: I have done something in between. Repeating the whole logic is troublesome for maintenance purposes as well, so the "main" logic is only written once, and the special case (whether to base62 or not) is moved to a function that basically is an identity operation on non windows environments, and base62 on windows.

flavio (approved):
LGTM, thanks for the changes!

viccuad (approved):
LGTM, yet looking forward to base64 (don't hesitate to re-request review).

ereslibre:
@viccuad I'm good to merge this now.

viccuad (approved):
LGTM! Thanks for the effort on this topic, and the function docs!

Issue #15, opened 2021-09-24
Do not download invalid WASM modules


ereslibre:
When an HTTP server or OCI registry does not report an error, and instead returns a `200` HTTP status when downloading the Wasm module, contents are downloaded blindly and put into the store.

This should not happen, and we should double check that what we downloaded is a valid Wasm module, and error out if it's not the case.

raulcabello:
This also happens when the http server returns a not found error `404`. Therefore users who made a typo in the url will experience this problem.

Steps to reproduce: 
- `kwctl pull https://github.com/kubewarden/pod-privileged-policy/releases/download/v0.1.9/policynotfound`
- Then `kwctl policies`                                                                                     
`Error: Bad magic number (at offset 0)`

PR #41, opened 2022-02-02
Validate wasm modules

raulcabello:
fixes https://github.com/kubewarden/policy-fetcher/issues/15


jvanz (approved):
LGTM! :D

viccuad (approved):
LGTM besides one comment.

src/lib.rs
```diff
@@ -194,6 +195,13 @@ pub(crate) fn host_and_port(url: &Url) -> Result<String> {
     ))
 }
 
+fn validate_wasm(buf: &[u8]) -> Result<()> {
+    match Module::validate(&Engine::default(), buf) {
```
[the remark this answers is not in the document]
raulcabello: I think this engine is already a wasmtime engine https://github.com/raulcabello/policy-fetcher/blob/d6b9319de47e32aaa9700810e3bab16aeaa47b8d/src/lib.rs#L32 or do you mean not use the default one?

src/https.rs
```diff
@@ -61,6 +62,7 @@ impl PolicyFetcher for Https {
 
         let client = client_builder.build()?;
         let buf = client.get(url.as_ref()).send().await?.bytes().await?;
+        validate_wasm(&buf)?;
```
ereslibre, line 65: I think it would be good to attempt a slightly different approach.

On https://github.com/kubewarden/policy-fetcher/blob/main/src/lib.rs#L40, when we call to `.fetch()`, we do so with a temporary file that we have created beforehand.

We inspect the temporary file. If it's not a valid Wasm module, we error out, otherwise we move it to the final destination and everything is fine.

I think it's a slightly better approach because the `validate_wasm` is not called by every backend but by the toplevel logic.

What do you think?

ereslibre (changes requested):
Thanks @raulcabello!

In general I think it is fine, but I think we can do an improvement to where the validation happens and how it happens.

src/https.rs
raulcabello: Thanks for the suggestion! I agree, it'd be better to call validate_wasm in the toplevel instead of calling it in each backend. But I don't find the temporary file that's created beforehand. That's why I made the call just before the file is created [here](https://github.com/raulcabello/policy-fetcher/blob/main/src/https.rs#L66) and [here](https://github.com/raulcabello/policy-fetcher/blob/main/src/registry/mod.rs#L276) so we avoid having to create the file and then delete it if there was an error. 

On a separate topic, I just noticed we are fetching the wasm file twice [here](https://github.com/raulcabello/policy-fetcher/blob/main/src/lib.rs#L97) and [here](https://github.com/raulcabello/policy-fetcher/blob/main/src/lib.rs#L110) (maybe 3 times [here](https://github.com/raulcabello/policy-fetcher/blob/main/src/lib.rs#L110) if we use tls too?) I put a println inside the https fetch method and I verified it's been called twice. I'll create an issue for this

ereslibre: Yes, this is in purpose. We behave as Docker does. The idea is:

1. We try to download the module with https and valid certificates
2. If the server is marked as insecure
    1. We try with https without checking certificates
    2. We try with http

This is akin to https://docs.docker.com/registry/insecure/

raulcabello: Thanks for the explanation, it makes more sense now. Anyway I still think we are calling it twice. Shouldn't we return OK if the first [fetch](https://github.com/raulcabello/policy-fetcher/blob/main/src/lib.rs#L97) is successful? As we don't return ok in the first fetch, the second will be tried and it will override the destination. 

Coming back to your original suggestion, I can't find the temporary file. That's why I make the call just before the file is created [here](https://github.com/raulcabello/policy-fetcher/blob/main/src/https.rs#L66) and [here](https://github.com/raulcabello/policy-fetcher/blob/main/src/registry/mod.rs#L276) so we avoid having to create the file and then delete it if there was an error. Where would you place the `validate_wasm` call in the toplevel logic? Sorry about diverting the original conversation

ereslibre: You are right, the first case is wrong and needs a match. The idea is to return early if succeeded, or if failed and is not an insecure registry. Otherwise, we continue with the rest of the retries.

> Coming back to your original suggestion, I can't find the temporary file

I explained myself extremely poorly. We are not doing this already, I was meaning to do that in this PR: create the temporary file, write the module to it, inspect the module. If the module is fine, we move it to the final location and succeed. If the module is wrong, we delete the temporary file and return error.

My idea was that the `File::create` in https://github.com/raulcabello/policy-fetcher/blob/d6b9319de47e32aaa9700810e3bab16aeaa47b8d/src/https.rs#L66 and https://github.com/raulcabello/policy-fetcher/blob/d6b9319de47e32aaa9700810e3bab16aeaa47b8d/src/registry/mod.rs#L277 is targeting the temporary file that has been created by https://github.com/raulcabello/policy-fetcher/blob/d6b9319de47e32aaa9700810e3bab16aeaa47b8d/src/lib.rs#L41 in advance.

So, summing up:

1. `fetch_policy` creates a temporary file, calls to the backend that matches with the temporary file path as the destination.
2. The backend does its thing and writes the contents to the destination, without knowing where that file is (is a temporary file, not a file in the real store, but the backend does not care).
3. The file is inspected back in `fetch_policy` or an utility function called by it.
4. If it's fine, the file is moved by `fetch_policy` or an utility function to the real place in the store and success is returned.
5. If it's corrupt, the file is removed by `fetch_policy` or an utility function and an error is reported back to the caller.

ereslibre: As an aside that I didn't answer. I think is fine to create a temporary file and discard it so we don't have to validate on every backend. Is true that not many backends will be needed and this might seem unnecessary, but I think doing the validation of the Wasm check belongs on an upper level where this logic is shared, instead of explicitly on each backend.

src/lib.rs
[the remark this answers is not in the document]
viccuad: doh, indeed! thanks for the pointer, disregard this.

src/fetcher.rs
```diff
@@ -1,6 +1,6 @@
 use anyhow::Result;
 use async_trait::async_trait;
-use std::path::Path;
+use bytes::Bytes;
```
ereslibre, line 3: I don't think we need `Bytes`. It's a data structure that implements some optimizations for working with the data directly. However, in our case, we just want to move whole byte streams around.

I would replace it with `Vec<u8>`, unless there is a strong reason not do so so.

src/https.rs
```diff
@@ -61,9 +54,6 @@ impl PolicyFetcher for Https {
 
         let client = client_builder.build()?;
         let buf = client.get(url.as_ref()).send().await?.bytes().await?;
```
ereslibre, line 56:
```suggestion
        client.get(url.as_ref()).send().await?.bytes().await
```

ereslibre: And you can get rid of `Ok(buf)`.

src/fetcher.rs
```diff
@@ -23,10 +23,5 @@ pub(crate) enum TlsVerificationMode {
 #[async_trait]
 pub(crate) trait PolicyFetcher {
     // Download the WASM module to the provided destination
```
ereslibre, line 25: Update comment

ereslibre (commented):
Thank you @raulcabello! Looking great! I love when a new feature is implemented and the code diff is net negative! Kudos!

Some comments.

src/fetcher.rs
raulcabello: done in https://github.com/kubewarden/policy-fetcher/pull/41/commits/75d7fa1c34c26d86b04e52fce9938e9d71f9d1c1

ereslibre (approved):
👏

src/https.rs
raulcabello: done in https://github.com/kubewarden/policy-fetcher/pull/41/commits/75d7fa1c34c26d86b04e52fce9938e9d71f9d1c1

src/fetcher.rs
raulcabello: done in https://github.com/kubewarden/policy-fetcher/pull/41/commits/75d7fa1c34c26d86b04e52fce9938e9d71f9d1c1

PR #42, opened 2022-02-04
Add verification config types from RFC

viccuad:
Currently, the `Subject` enum types use both `#[serde(deny_unknown_fields)]` and `#[serde(flatten)] `, which is not supported, and is supposed to fail on deserialization. Strangely, it works here, as the tests pass.

Also tested by consuming `verify::read_verification_file()` from a simple crate, instead of consuming this branch from policy-server and policy-evaluator.

Part of https://github.com/kubewarden/policy-server/issues/142.

PR #54, opened 2022-02-21
implement policy verification code

flavio:
>**Warning:** this PR builds on top of https://github.com/kubewarden/policy-fetcher/pull/50 and supersedes both https://github.com/kubewarden/policy-fetcher/pull/50 and https://github.com/kubewarden/policy-fetcher/pull/42

The following changes are done to the code submitted inside of https://github.com/kubewarden/policy-fetcher/pull/50:

- Update to latest version of sigstore: some breaking API changes got merged upstream, I had to bring in them in right now because they broke how the built the public key constraint verifier.
- Add more verification constraints: the `todo()!` blocks of PR #50 have been addressed. Now all the constraints described inside of our RFC are implemented and tested.
- Simplify the verification settings structs. Get rid of the `flatten` directive and of the `UrlIssuer`, more on that below.
- DRY-ing the code: remove some code repetition, as suggested during the review process of #50

As a result of this PR, the `verification.yml` file now looks like that:

```yaml
apiVersion: v1
allOf:
  - kind: genericIssuer
    issuer: https://token.actions.githubusercontent.com
    subject:
       equal: https://github.com/kubewarden/policy-secure-pod-images/.github/workflows/release.yml@refs/heads/main
  - kind: genericIssuer
    issuer: https://token.actions.githubusercontent.com
    subject:
       urlPrefix: https://github.com/kubewarden
```

As you can see, the `subject` attribute has now a nested element which could be either `equal` or `urlPrefix`. This is slightly different compared to the RFC, but it allows us to get rid of the `flatten` feature of serde which doesn't play well with teh `deny_unknown_fields` one.
I'll update the RFC once this gets merged.


src/verify/verification_constraints.rs
```diff
@@ -0,0 +1,628 @@
...
+    use sigstore::{cosign::signature_layers::CertificateSignature, simple_signing::SimpleSigning};
+
+    fn build_signature_layers_pub_key<'a>() -> (&'a str, SignatureLayer) {
+        let pub_key: &'a str = r#"-----BEGIN PUBLIC KEY-----
+MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAELKhD7F5OKy77Z582Y6h0u1J3GNA+
+kvUsh4eKpd1lwkDAzfFDs7yXEExsEkPPuiQJBelDT68n7PDIWB/QEY7mrA==
+-----END PUBLIC KEY-----"#;
+        let raw_data = r#"{"critical":{"identity":{"docker-reference":"registry-testing.svc.lan/busybox"},"image":{"docker-manifest-digest":"sha256:f3cfc9d0dbf931d3db4685ec659b7ac68e2a578219da4aae65427886e649b06b"},"type":"cosign container image signature"},"optional":null}"#;
+        let raw_data = raw_data.as_bytes().to_vec();
+        let signature = "MEYCIQDWWxPQa3XFUsPbyTY+n+bZu/6Pwhg5WwyYDQtEfQho9wIhAPkKW7eub8b7BX+YbbRac8TwwIrK5KxvdtQ6NuoD+ivW".to_string();
```
viccuad, line 264: We could add a comment with the incantation to create the signature.

src/verify/verification_constraints.rs
```diff
@@ -0,0 +1,628 @@
...
+        });
+
+        let s = match &subject {
+            Subject::Equal(_) => subject.clone(),
+            Subject::UrlPrefix(url) => {
+                let prefix = url.to_string();
+                if prefix.ends_with('/') {
+                    subject.clone()
+                } else {
+                    let u = url::Url::parse(format!("{}/", prefix).as_str())
```
viccuad, line 81: This should never happen if deserializing with policy-fetcher, given that we append `/` when deserializing (and have tests). But good to have it, plus its tests too.

viccuad (approved):
LGTM, thanks for the changes!

The verification config has changes, with:
```yaml
- kind: urlIssuer
  url: http://foo.com
```
being substituted for:
```yaml
- kind: genericIssuer
  issuer: http://foo.com
```

Which means that URLs in issuer don't get deserialized and validated as URLs. I'm ok with it though.

I like the approach of `subject.equal` and `subject.urlprefix` in the config yaml. Once this is merged we should update the RFC with the changes.

src/verify/verification_constraints.rs
flavio: yes, I added it especially for the unit tests, where we create things in a more direct way :+1:

src/verify/verification_constraints.rs
flavio: done

flavio:
@viccuad I've applied the changes you requested and I have rebased against `main`. I would go ahead and merge everything as it is.

Are you fine with that?

viccuad:
sounds perfect, thanks! Merging then :).

PR #55, opened 2022-02-24
Add verification algorithm tests, and fix failures

viccuad:
## Description

Relates to https://github.com/kubewarden/policy-server/issues/142.

Depends on https://github.com/sigstore/sigstore-rs/pull/40.

## Test

Added tests as part of unit tests, run with `make test`.

### Tradeoff

Refactored the verification algorithm, which filters with contraints, into its own function.

### Potential improvement

Another option could have been to add integration tests under `tests/`. But that entails mocking the registry and image, with its layers, and the codepath under test wouldn't have grown much, as sigstore-rs has its own tests.

### DIFF
--- a/src/verify/config.rs
+++ b/src/verify/config.rs
@@ -114,6 +114,11 @@ where
 pub fn read_verification_file(path: &Path) -> Result<VerificationSettings> {
     let settings_file = File::open(path)?;
     let vs: VerificationSettings = serde_yaml::from_reader(&settings_file)?;
+    if vs.all_of.is_none() && vs.any_of.is_none() {
+        return Err(anyhow!(
+            "config is missing signatures in both allOf and anyOff list"
+        ));
+    }
     Ok(vs)
 }
 
--- a/src/verify/config.rs
+++ b/src/verify/config.rs
@@ -11,7 +11,7 @@ use crate::verify::verification_constraints;
 
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
 #[serde(rename_all = "camelCase", deny_unknown_fields)]
-pub struct VerificationSettings {
+pub struct VerificationConfig {
     pub api_version: String,
     pub all_of: Option<Vec<Signature>>,
     pub any_of: Option<AnyOf>,
@@ -112,15 +112,15 @@ where
     Ok(url)
 }
 
-pub fn read_verification_file(path: &Path) -> Result<VerificationSettings> {
-    let settings_file = File::open(path)?;
-    let vs: VerificationSettings = serde_yaml::from_reader(&settings_file)?;
-    if vs.all_of.is_none() && vs.any_of.is_none() {
+pub fn read_verification_file(path: &Path) -> Result<VerificationConfig> {
+    let config_file = File::open(path)?;
+    let config: VerificationConfig = serde_yaml::from_reader(&config_file)?;
+    if config.all_of.is_none() && config.any_of.is_none() {
         return Err(anyhow!(
             "config is missing signatures in both allOf and anyOff list"
         ));
     }
-    Ok(vs)
+    Ok(config)
 }
 
 #[cfg(test)]
@@ -140,7 +140,7 @@ mod tests {
         #subject:
         #   urlPrefix: https://github.com/kubewarden/
     "#;
-        let _vs: VerificationSettings = serde_yaml::from_str(config).unwrap();
+        let _vs: VerificationConfig = serde_yaml::from_str(config).unwrap();
     }
 
     #[test]
@@ -159,7 +159,7 @@ mod tests {
            urlPrefix: https://github.com/kubewarden
     "#;
 
-        let vs: VerificationSettings = serde_yaml::from_str(config).unwrap();
+        let vs: VerificationConfig = serde_yaml::from_str(config).unwrap();
         let signatures: Vec<Signature> = vec![
             Signature::GenericIssuer {
                     issuer: "https://token.actions.githubusercontent.com".to_string(),
@@ -172,7 +172,7 @@ mod tests {
                 annotations: None,
             }
         ];
-        let expected: VerificationSettings = VerificationSettings {
+        let expected: VerificationConfig = VerificationConfig {
             api_version: "v1".to_string(),
             all_of: Some(signatures),
             any_of: None,
@@ -195,7 +195,7 @@ mod tests {
         subject:
            urlPrefix: https://github.com/kubewarden/ # should deserialize path to kubewarden/
     "#;
-        let vs: VerificationSettings = serde_yaml::from_str(config).unwrap();
+        let vs: VerificationConfig = serde_yaml::from_str(config).unwrap();
         let signatures: Vec<Signature> = vec![
             Signature::GenericIssuer {
                 issuer: "https://token.actions.githubusercontent.com".to_string(),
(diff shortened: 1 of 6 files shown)


viccuad:
Leaving it as draft until I can consume https://github.com/sigstore/sigstore-rs/pull/40.

src/verify/config.rs
```diff
@@ -114,6 +114,11 @@ where
 pub fn read_verification_file(path: &Path) -> Result<VerificationSettings> {
     let settings_file = File::open(path)?;
     let vs: VerificationSettings = serde_yaml::from_reader(&settings_file)?;
+    if vs.all_of.is_none() && vs.any_of.is_none() {
```
viccuad, line 117: this looks like it should be in a builder pattern of `VerificationSettings`. Maybe a builder where one passes `version`, `all_of` and `any_of` as underlying structs (Signature, Subject) need to be kept public to simplify..

src/verify/mod.rs
```diff
@@ -229,30 +229,35 @@ fn verify_signatures_against_settings(
...
-                "Image verification failed: no matching signature found on AllOf list"
-            ))
-        }
-        Err(e) => return Err(anyhow!("{}", e)),
-        Ok(m) if m.len() <= length_constraints_all_of => {
-            return Err(anyhow!(
-                "Image verification failed: missing signatures in AllOf list"
-            ));
+    if verification_settings.all_of.is_none() && verification_settings.any_of.is_none() {
+        // deserialized config is already sanitized, and should not reach here anyways
```
viccuad, line 233: this should be `unreachable` if we have a `VerificationSettings` builder.

src/verify/config.rs
viccuad: On second thought, given that `verify_signatures_against_settings()` checks for this already by using `vs.all_of`, `vs.any_of`, there's no need now. Maybe in the future we can add the builder for version, and getters for `vs.all_of` and `vs.any_of`.

src/verify/mod.rs
```diff
@@ -244,3 +203,280 @@ impl Verifier {
         }
     }
 }
+
+// Verifies the trusted layers against the verification settings passed to it.
+// It does that by creating the verification constraints from the settings, and
+// then filtering the trusted_layers with the corresponding constraints.
+fn verify_signatures_against_settings(
```
viccuad, line 210: Not for now, but this function should print debug/warning outputs with which signatures are not verified, instead of a blanket "missing signatures".

For this, we need to list those `SignatureLayer`s that don't pass any verification constraint: those `SignatureLayer`s that aren't present in `m` from `cosign::filter_signature_layers()`.

We can iterate for each `SignatureLayer` on the image, which means `O(L*m)`, : `L` being number of layers in image, `m` being layers that are verified already by the constraints.
This can be parallelized with the [rayon crate](https://crates.io/crates/rayon), giving us `O(L)`. 
Another approach would be to create a HashMap with the L layers, to reduce the complexity to `O(L)`. But then we need to assure that SignatureLayers are unique.

viccuad: Punted and opened https://github.com/kubewarden/policy-fetcher/issues/57.

viccuad:
Force-pushed to change the error strings and remove mentions to "AllOf" and "AnyOf". Since consumers such as kwctl build a VerificationSettings from flags, and users may not know about "AllOf" or "AnyOf" in that context.

This is, as it was before, ready for review.

viccuad (commented):
Now that we are starting to consume this, we should maybe rename `src/verify/config.rs`, `config` module, to `src/verify/config-v1.rs` and `config-v1` module?

ereslibre:
> Now that we are starting to consume this, we should maybe rename src/verify/config.rs, config module, to src/verify/config-v1.rs and config-v1 module?

I think I would favour `src/verify/config.rs` with versioned configurations as nested types (e.g. `Verify::Config::Versions::V1`, while the toplevel `Verify::Config` is an alias of the latest supported one.

When more versions are supported, `Verify::Config` could have some logic to `TryInto<Verify::Versions::V1>` or `TryInto<Verify::Versions::V2>` while the API only exposes `Verify::Config` in their arguments -- meaning: API users only use the "external" representation, and the crate knows how to convert from/into "internal" representations (`Versions::V1`, `Versions::V2`...)

src/verify/mod.rs
```diff
@@ -244,3 +205,274 @@ impl Verifier {
...
+            "Image verification failed: no signatures to verify"
+        ));
+    }
+
+    if verification_settings.all_of.is_some() {
+        let length_constraints_all_of = constraints_all_of.len();
+        match cosign::filter_signature_layers(trusted_layers, constraints_all_of) {
+            Ok(verified_layers) if verified_layers.len() < length_constraints_all_of => {
+                return Err(anyhow!("Image verification failed: missing signatures"));
+            }
```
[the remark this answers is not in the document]
viccuad: That would mean that we expect all layers of the image to be verified, which may not be true (we can have an image with 10 layers, we filter with 2 constraints in allOf, and we only need 2 verified layer to pass verification, not the whole 10).

But this is all moot, because we can't use `trusted_layers.len()` nor `constraints.len()` for seeing if we are passing verification. A constraint could make several layers pass at once, as applying constraints is non an injective function. We are lucky right now that all constraints implemented in policy-fetcher gives us an injective 1-to-1 function, but if we would expose and use the AnnotationsConstraint from sigstore, that wouldn't be the case. 

What we really care about is filtering the constraints into those that verified, and those that not, instead of filtering the layers into those that pass verification, and those that not (see https://github.com/kubewarden/policy-fetcher/issues/57).

I propose the implement https://github.com/kubewarden/policy-fetcher/issues/57, and return also a vectors of constraints that fail. This is needed anyways to provide meaningful logs to users.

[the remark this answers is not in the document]
viccuad: Opened https://togithub.com/sigstore/sigstore-rs/issues/41 to work on it.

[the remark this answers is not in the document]
flavio: You're right. I wonder if we couldn't simply fix that by changing the usage of `any` to [`all`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.all) inside of [this](https://github.com/sigstore/sigstore-rs/blob/9b561afb03cbd21f5960cf5628e7e99354f5e3c2/src/cosign/mod.rs#L104) code.

Surely knowing which constraints got rejected would be useful for the end user. I think we should extend the verification constraints to implement also something like the `Display` trait. Otherwise it would  be problematic for the end users to debug it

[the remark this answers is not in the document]
viccuad: Resolving, this is now fixed upstream.

viccuad:
Rebased onto main. I would like to solve `Config` alias before merging though, so consumers don't need to change twice.

viccuad:
This is ready for review.

src/verify/verification_constraints.rs
```diff
@@ -46,6 +51,12 @@ impl VerificationConstraint for PublicKeyAndAnnotationsVerifier {
         } else {
             self.pub_key_verifier.verify(sl)?
         };
+        if !outcome && self.owner.is_some() {
+            info!(
+                "pubkey not satisfied for owner {}",
+                self.owner.as_ref().unwrap()
+            );
```
flavio, line 58: I would always print this message, regardless of `owner` being something.

What about:

```rust
info!( owner ?= owner, "pubkey not satisfied");
```

flavio (approved):
LGTM, I left a minor comment. Once this is addressed (if you feel like this should be addressed), feel free to merge the PR

src/verify/verification_constraints.rs
viccuad: Yes, it's indeed better to always print the info. Fixed on https://github.com/kubewarden/policy-fetcher/pull/55/commits/0ce0ec18163af8ddc92f3b07394501632a0f6260.

Issue #59, opened 2022-03-08
Feature Request:  Provide versioned and latest types for `VerificationConfig`


viccuad:
### Is your feature request related to a problem?

From https://github.com/kubewarden/policy-fetcher/pull/55#issuecomment-1055527972, right now, `VerificationConfig` has a hardcoded version. If we were to bump the version and change the `VerificationConfig`struct, or any struct inside of it, it will break for consumers (given that we set `deny_unknown_fields`).

### Solution you'd like

Provide an aliased type, `VerificationConfigLatest` or similar, that maps to the latest config.
Provide versioned types, `VerificationConfigV1`, that implements the current type.
Provide an enum that maps the latest version to the different versions. This enum can implement `TryInto` in the future, to seamlessly cast versions to the latest.

Consumers will only consume `VerificationConfigLatest`.

flavio:
Implemented

PR #60, opened 2022-03-09
Implement better versioning of VerificationConfig

flavio:
Introduce a series of types that allow a better versioning of the verification config.

This fixes issue #59

### DIFF
--- a/src/verify/config.rs
+++ b/src/verify/config.rs
@@ -4,7 +4,7 @@ use sigstore::{
     cosign::verification_constraint::VerificationConstraint, crypto::SignatureDigestAlgorithm,
 };
 use std::boxed::Box;
-use std::{collections::HashMap, fs::File, path::Path};
+use std::{collections::HashMap, fs, path::Path};
 use url::Url;
 
 use crate::verify::verification_constraints;
@@ -43,7 +43,7 @@ pub enum VersionedVerificationConfig {
 #[serde(untagged)]
 pub enum VerificationConfig {
     Versioned(VersionedVerificationConfig),
-    Invalid(serde_json::Value),
+    Invalid(serde_yaml::Value),
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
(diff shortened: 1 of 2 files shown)


src/verify/config.rs
```diff
@@ -9,14 +9,43 @@ use url::Url;
...
+    Unsupported,
+}
+
+/// Enum that distinguish between a well formed (but maybe unknown) version of
+/// the verification config, and something which is "just wrong".
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum VerificationConfig {
+    Versioned(VersionedVerificationConfig),
+    Invalid(serde_json::Value),
```
viccuad, line 46: should we use `serde_yaml`?

src/verify/config.rs
```diff
@@ -112,9 +141,25 @@ where
...
+            VersionedVerificationConfig::V1(c) => c,
+            VersionedVerificationConfig::Unsupported => {
+                return Err(anyhow!(
+                    "Not a supported configuration version: {:?}",
+                    versioned_config
+                ))
+            }
+        },
+        VerificationConfig::Invalid(value) => {
+            return Err(anyhow!("Not a valid configuration file: {:?}", value))
```
viccuad, line 159: IIUC this prints the invalid config back to the user. I would be more interested on the errors on why it did serialize incorrectly.

Also, now `read_verification_file()` has become more complex, maybe a test on it checking for those errors would be good.

src/verify/config.rs
```diff
@@ -140,7 +184,8 @@ mod tests {
         #subject:
         #   urlPrefix: https://github.com/kubewarden/
     "#;
-        let _vs: VerificationConfig = serde_yaml::from_str(config).unwrap();
+        let vc: VerificationConfig = serde_yaml::from_str(config).unwrap();
```
viccuad, line 187: Now we can deserialize into `VerificationConfig::Invalid`. Which means that users lose the errors like `"missing field `subject`"`, which were useful.

src/verify/config.rs
flavio: Good point, it still worked... LOL

I'll fix it

src/verify/config.rs
flavio: This error happens when you have a valid YAML file (from a syntax POV), but the contents do not make any sense to us.

Something like:
```yaml
grocery store:
- item: milk
  qty: 1L
- item: bread
  qty: 500g
- item: cereals
  flavor: chocolate
  qty: 2
```

I'll extend a bit the test  case :+1:

src/verify/config.rs
flavio: I fear there's nothing we can do to address that, I'll look into that. I fear we will have to have a compromise

flavio:
@viccuad please take another look at the PR, I've addressed your concerns with the latest commit

viccuad (approved):
LGTM, thanks!

ereslibre (approved):

PR #61, opened 2022-03-10
Remove wasmtime dependency

flavio:
The wasmtime dependency was introduced to have a reliable way to detect corrupted Wasm modules.
That was done because GitHub Container Registry sometimes returns a html page instead of a Wasm module.

The wasmtime dependency is a bit of an overkill, plus it makes maintenance a bit harder because it could cause dependency resolution conflicts between policy-fetcher and other kubewarden crates that **need** wasmtime.

This commit removes the wasmtime dependency and instead identifies valid Wasm files by looking at their ["magic bytes"](https://en.wikipedia.org/wiki/List_of_file_signatures).


raulcabello (approved):
LGTM, it's great that we can remove the wasmtime dependency

src/lib.rs
```diff
@@ -183,9 +182,16 @@ pub(crate) fn host_and_port(url: &Url) -> Result<String> {
     ))
 }
 
-fn create_file_if_valid(bytes: Vec<u8>, destination: &Path, url: String) -> Result<Policy> {
-    if let Err(err) = Module::validate(&Engine::default(), bytes.as_ref()) {
-        return Err(anyhow!("invalid wasm file: {}", err));
+// Each Wasm file begins with a well known bytes sequence, known as
+// "magic bytes" (see https://en.wikipedia.org/wiki/List_of_file_signatures).
+//
+// The Wasm magic bytes sequece is defined inside of its official specification:
```
ereslibre, line 188:
```suggestion
// The Wasm magic bytes sequence is defined inside of its official specification:
```

tests/test_data/README.md
```diff
@@ -0,0 +1,18 @@
+Fixtures used inside of unit tests.
+
+## `simple.wasm`
+
+This is the simples Wasm module that can be produced while still being compliant
```
ereslibre, line 5:
```suggestion
This is the simplest Wasm module that can be produced while still being compliant
```

tests/test_data/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+```
+(module)
+```
+
+and then invoke:
+
+```console
+wat2wasm simple.wat
+```
```
ereslibre, line 18: Kudos for documenting how to generate a binary blob used in tests!

ereslibre (approved):
This is great. Thanks! Just a couple of typos.

viccuad (approved):
LGTM, thanks for this!

PR #67, opened 2022-03-14
verifier: improve how fulcio and rekor data is found

flavio:
Allow consumers of the library to either provide fulcio and rekor data manually, or by fetching it from a TUF repository.

The crate is now re-exposing the `sigstore` and the `oci-distribution` crates. This is required to simplify dependency management, because these are 2 crates that can be used by consumers of this library.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -29,6 +29,10 @@ use std::path::{Path, PathBuf};
 use tracing::debug;
 use url::ParseError;
 
+// re-export for usage by kwctl, policy-server, policy-evaluator,...
+pub use oci_distribution;
+pub use sigstore;
+
 #[derive(Debug)]
 pub enum PullDestination {
     MainStore,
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -29,6 +29,10 @@ use std::path::{Path, PathBuf};
 use tracing::debug;
 use url::ParseError;
 
+// re-export for usage by kwctl, policy-server, policy-evaluator,...
+pub use oci_distribution;
```
viccuad, line 33: I see re-exporing `sisgtore`, but why `oci_distribution`, just for future proofing?

viccuad (approved):
LGTM, just a question.

jvanz (approved):

src/lib.rs
flavio: Yes, I intend to clean up policy-evaluator, which is requiring inside of its `Cargo.toml` both `policy-fetcher` and `oci-distribution`.

PR #69, opened 2022-03-15
Changes to the verification API

flavio:
Small changes to the verification API to use more references to objects. 

These changes are needed to make the policy-server code better (avoid lots of `clone()` invocations).

I've already tested these changes with kwctl, policy-server, policy-evaluator. Everything works fine, I'm pretty sure I'll not need more changes :sweat_smile: 

There's just one last question: we will have to tag a new release. Given these are breaking API changes, should we make it `v0.6.0` or should we go with
`v0.5.1`? I'm bringing up the possibility to have a patch release because `v0.5.0` was tagged just some hours ago...

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "policy-fetcher"
-version = "0.5.0"
+version = "0.6.0"
 authors = [
   "Flavio Castelli <fcastelli@suse.com>",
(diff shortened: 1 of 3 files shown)


viccuad (approved):
LGTM, agree with 0.6.0.

flavio:
I've bumped it to 0.6.0 :+1:

PR #71, opened 2022-03-18
Derive `Default` for `VerificationConfigV1`

ereslibre:
Related: https://github.com/kubewarden/policy-server/issues/220

### DIFF
--- a/src/verify/config.rs
+++ b/src/verify/config.rs
@@ -17,7 +17,7 @@ use crate::verify::verification_constraints;
 /// * Implement `TryFrom` that goes from (v - 1) to (v)
 pub type LatestVerificationConfig = VerificationConfigV1;
 
-#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


ereslibre:
Dropping this one.

viccuad (approved):
Approving to get the fix going.

But I would very much prefer for consumers to not be able to create empty VerificationConfigs with `default()`; Right now the algorithm will correctly fail verification in those cases, but to me it's an antipattern.

ereslibre:
> But I would very much prefer for consumers to not be able to create empty VerificationConfigs with default(); Right now the algorithm will correctly fail verification in those cases, but to me it's an antipattern.

I'm working to make this PR unnecessary. I wonder though why an `all_of: None, any_of: None` is an antipattern as the default. I would expect anything being verified to pass (the opposite of what it does not) against `None` in both.

In any case, that is a different discussion on how the verifier should behave when `all_of: None, any_of: None` than the discussion of if it makes sense to implement a default on this type.

ereslibre:
With the last update of the PR https://github.com/kubewarden/policy-server/pull/223 this is no longer necessary. Closing.

PR #73, opened 2022-03-21
Parallelization of the signatures verifications  . 

### LINKED ISSUE #65: Feature Request: Paralelize verification of constraints
### Is your feature request related to a problem?

_No response_

### Solution you'd like

With https://github.com/kubewarden/policy-fetcher/pull/64, now it is easier to parallelize iterating through all the constraints, as we have our own iter: https://github.com/viccuad/policy-fetcher/blob/a79ff548b2e4be912b746ae6799d35d1e6642b00/src/verify/mod.rs#L228.

This function could be parallelized with async + await. Given that the `VerificationConstraint`s are Box<dyn>, we may need to implement Send and Sync for them.

Another possible option may be to use [rayon](https://docs.rs/rayon/latest/rayon/) and `par_iter()`.

### Alternatives you've considered

_No response_

### Anything else?

_No response_

jvanz:
## Description

Updates the function used to verify if the signatures are respected to check multiple signatures in parallel. For this, it's used the stream module from async_std crate.

Fix #65

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ sigstore = { git = "https://github.com/sigstore/sigstore-rs", default-features =
 tracing = "0.1.32"
 url = { version = "2.2.2", features = ["serde"] }
 walkdir = "2"
+rayon = "1.5.1"
 
 [dev-dependencies]
 rstest = "0.12.0"
--- a/src/verify/mod.rs
+++ b/src/verify/mod.rs
@@ -274,9 +274,11 @@ fn verify_signatures_against_config(
         ));
     }
 
+    use rayon::prelude::*;
+
     if let Some(ref signatures_all_of) = verification_config.all_of {
         let unsatisfied_signatures: Vec<&Signature> = signatures_all_of
-            .iter()
+            .par_iter()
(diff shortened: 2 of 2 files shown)


jvanz:
I'm marking this PR as draft because I wanna refactor the code removing duplicated code. However, any feedback is welcome.

flavio (commented):
Can you try to implement that with [rayon](https://crates.io/crates/rayon)? It has more adoption than async-std and, most important of all, seems to require less changes

jvanz:
> Can you try to implement that with [rayon](https://crates.io/crates/rayon)? It has more adoption than async-std and, most important of all, seems to require less changes

No problems. I've decided to use async-std because it is already in our dependencies list and avoid another dependency. But I agree that rayon seems better. :)

raulcabello (approved):
LGTM. rayon is really cool!

viccuad (approved):
LGTM! Thanks for the previous try with async!

ereslibre (approved):
Super cool 👏

jvanz:
> LGTM. rayon is really cool!

YES! Simple and elegant!

PR #77, opened 2022-04-04
feat: Allow policy authors to write policies that perform image verif…

raulcabello:
Move sigstore verification enums to the sdk

Fix https://github.com/kubewarden/policy-server/issues/99

PR #79, opened 2022-04-14
feat: Add pub `verify_pub_keys()`, `verify_keyless_exact_match()`

viccuad:
## Description

Fix https://github.com/kubewarden/policy-server/issues/99
Supersedes https://github.com/kubewarden/policy-fetcher/pull/77

- Add pub functions `verify_pub_keys()`, `verify_keyless_exact_match()`. These are shims that translate to an interim `VerificationConfig` and call `verify()`, so only 1 codepath exists, no need for tests.
- re-export kubewarden_policy_sdk through policy-fetcher for consumption.
- Minor refactors and error checking.

I suggest "files-changed" review, given the hard refactors.

TODO: depends on rust sdk work.

### DIFF
--- a/src/verify/mod.rs
+++ b/src/verify/mod.rs
@@ -4,7 +4,10 @@ use crate::{policy::Policy, registry::config::DockerConfig};
 use anyhow::{anyhow, Result};
 use oci_distribution::manifest::WASM_LAYER_MEDIA_TYPE;
 use sigstore::cosign::{self, signature_layers::SignatureLayer, ClientBuilder, CosignCapabilities};
+use std::collections::HashMap;
 
+use oci_distribution::Reference;
+use std::convert::TryFrom;
 use std::{convert::TryInto, str::FromStr};
 use tracing::{debug, error, info};
 use url::{ParseError, Url};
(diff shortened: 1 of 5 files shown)


flavio:
Ok, I looked into the PR and I understand your comment. I didn't expect the code to reuse the verification primitives of policy-fetcher. Given the limited API offered to the end user, I was expecting to see something written on top of "pure sigstore", plus some little duplicated code. I understand why you went this way.

At the same time, I think it's just a bit strange to me to have the `policy-sdk` requirement be moved from `policy-evaluator` to `policy-fetcher`. Moreover it's strange for me to look for the implementation details of an exposed callback function not into `policy-evaluator`, but `policy-fetcher`.
I also think it would be annoying to implement fixes to the verification APIs, because would have to touch `policy-fetcher` and then `policy-evaluator`.

Being pragmatic, for the end user there's no difference. This is definitely a matter of personal taste. I'm fine leaving the code as it currently is. We can always change that in the future without breaking user policies.

raulcabello:
We are sharing some code with the policy validation such as `FulcioAndRekorData` and `cosign client` creation in `Verifier`. We could move the two methods `verify_pub_keys()` and  `verify_keyless_exact_match() `to the evaluator and still reuse the verify code in the `Verifier`.

However,  I think it conceptually makes more sense to have all verify functions together in the `Verifier` struct in `policy-fetcher`. This would help with refactors or changes in VerificationConfig as @viccuad pointed out.

flavio (approved):
LGTM, as I stated I would have moved the code into `policy-evalutor`, but this is a matter of personal taste

viccuad (commented):
Raul's new changes LGTM, thanks! and thanks for the sigstore registry change, forgot about it.

viccuad:
Still, now we don't need to re-export the SDK through policy-fetcher, nor consume it in evaluator.

raulcabello:
@viccuad we still needed [here](https://github.com/raulcabello/policy-evaluator/blob/sigstore_container_verification3/src/callback_handler/mod.rs#L10-L13) for the `CallbackRequestType`, `KeylessInfo` and `VerificationResponse`

viccuad:
Rebased main, and consumed policy-sdk-rust from Raúl's branch, of the other PR, for now.

src/verify/mod.rs
```diff
@@ -97,49 +99,91 @@ impl Verifier {
...
+            };
+            signatures_all_of.push(signature);
+        }
+        let verification_config = LatestVerificationConfig {
+            all_of: Some(signatures_all_of),
+            any_of: None,
+        };
+        self.verify(&image_url, docker_config, &verification_config)
+            .await
+    }
```
flavio, line 153: This method is causes the `policy-sdk` dependency to be inside of `policy-fetcher`. Given this method leverages the public `verify` method, can't we simply move it into `policy-evaluator`?

In that way we will get rid of the `policy-sdk` dependency

flavio (changes requested):

src/verify/mod.rs
raulcabello: that was done, but I can't see that change here anymore. I still see the commit in my local though..

raulcabello: it was done [here](https://github.com/kubewarden/policy-fetcher/commit/a1f46afdefc0c1de9958379373cd26d3a029223a)

flavio: There's something wrong with the PR then, when I open your link I get a big warning at the top of it:

> **This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository.**

flavio (approved):
LGTM, I love how small this change turned out to be

src/verify/mod.rs
```diff
@@ -97,49 +99,38 @@ impl Verifier {
...
-    /// specified inside of the Sigstore signature object.
-    ///
-    /// In case of success, returns the manifest digest of the verified policy.
-    ///
-    /// Note well: this method doesn't compare the checksum of a possible local
-    /// file with the one inside of the signed (and verified) manifest, as that
-    /// can only be done with certainty after pulling the policy.
-    ///
-    /// Note well: right now, verification can be done only against policies
-    /// that are stored inside of OCI registries.
```
viccuad, line 113: this commentary has disappeared.

viccuad: Reinstated in https://github.com/kubewarden/policy-fetcher/pull/79/commits/3ead76d6c18f91e3e1160a1fca7781693dbdc145.

flavio (approved):

jvanz (approved):

PR #111, opened 2022-08-02
fix: AWS ECR Authorization not recognized after Docker Login

raulcabello:
## Description

Currently we are just using plain text credentials stored in the `docker/config.json`. We are ignoring if a credentials store is used, which is considered best practice and the default behavior if you are using Docker Desktop. 

In this PR it will always try to first fetch credentials from credentials store if provided. Fallback to auth credentials in plain text if  `docker/config.json` does not contain a credentials store.

Edit: Use https://github.com/keirlawson/docker_credential which will fetch credentials from credentials store or helper

Fix https://github.com/kubewarden/kwctl/issues/208

## Test

you can use these binaries for testing in windows and mac https://github.com/raulcabello/kwctl/releases/tag/v0.9-test-credstore


src/registry/config.rs
```diff
@@ -50,6 +62,41 @@ impl DockerConfig {
...
+    ) -> Option<Result<RegistryAuth>> {
+        self.creds_store
+            .as_ref()
+            .map(|creds_store| get_auth_from_credentials_helper(creds_store.as_str(), registry))
+    }
+}
+
+fn get_auth_from_credentials_helper(creds_store: &str, registry: &str) -> Result<RegistryAuth> {
+    let mut process = Command::new(format!("docker-credential-{}", creds_store))
+        .arg("get")
```
viccuad, line 82: Do the  `docker-credential-*` binaries  have a defined UX API? Looked around, couldn't see one.
I suppose having the `get` is ok, but I wonder if this may break in unsuspected ways for some users of some cred stores. The same with declaring a cred store, but missing the accompanying CLI binary.

viccuad: Maybe it is worth it to delegate all of this to https://docs.rs/crate/docker_credential/latest?

src/registry/config.rs
```diff
@@ -17,8 +19,10 @@ pub(crate) struct RegistryAuthRaw {
 }
 
 #[derive(Deserialize, Debug)]
+#[serde(rename_all = "camelCase")]
```
jvanz, line 22: Just double check: you're using [different values for](https://github.com/kubewarden/policy-fetcher/pull/111/files#diff-cf3e7af3d3a6ef1b9451d6c5ffb44c211234fc152e9b149289ade8fd6027aeafR46) the `rename_all` attribute. Is it what you want to do?

src/registry/config.rs
jvanz: I agree with @viccuad. I would prefer to use a crate to parse the file. However, I'm not sure if the `docker_credential` crate is a good idea. It seems unmaintained to me. Maybe we can help maintain it or fork it.

flavio: > Do the `docker-credential-*` binaries have a defined UX API? Looked around, couldn't see one. I suppose having the `get` is ok, but I wonder if this may break in unsuspected ways for some users of some cred stores. The same with declaring a cred store, but missing the accompanying CLI binary.

This doesn't seem to be super documented, but the actual cli interface is defined here:

https://github.com/docker/docker-credential-helpers/blob/e595cd69465c6b0f7af2d49582b82fdeddecbf75/credentials/credentials.go#L51-L65

raulcabello: You can find the API [here](https://docs.docker.com/engine/reference/commandline/login/) in the `Credential helper protocol` section. API should be the same for all  `docker-credential-*` binaries.
@viccuad I'd love to use an existing crate for this, but as @jvanz pointed out this crate seems unmaintained and the latest release is more than 3 years old.
I just noticed this PR is implementing just credsStore, which is the problem for the bug we are fixing. There is also credentials helpers:

> Credential helpers are similar to the credential store above, but act as the designated programs to handle credentials for specific registries. The default credential store (credsStore or the config file itself) will not be used for operations concerning credentials of the specified registries.

Should I add credential helpers in this PR? or create a separate issue?

src/registry/config.rs
raulcabello: yes, that is what we want. `DockerConfigRaw` is reading the `.docker/config.json` that is using camelCase. And the response from the credentials helper is PascalCase

src/registry/config.rs
raulcabello: It is easy to add, so I'll add it to this PR. This is the priority of methods for authenticating.
1. look for credentials helpers, which are registry specific (not implemented yet, will add it now)
2. look for credsStore (implemented in this PR and was the cause for the bug reported)
3. use plain text credentials (what we had before)
4. anonymous

src/registry/config.rs
```diff
@@ -50,6 +62,45 @@ impl DockerConfig {
 
         Ok(self.auths.get(reference.registry()).cloned())
     }
+
+    pub fn get_auth_from_credentials_helper_if_present(
```
flavio, line 66: nit, adding some documentation about what the method does would be nice.

Something has short as:

```suggestion
    /// Get registry credentials using an external credential-helper.
    /// Returns `None` when no credential-helper is configured for the requested registry
    /// Otherwise returns `Some`-thing when a credential-helper is configured. The inner object
    /// can be an `Err` if something went wrong while invoking the external helper
    pub fn get_auth_from_credentials_helper_if_present(
```

flavio (commented):
The code looks good to me. The changes are small and the tests are nice.

I have mixed feelings, because the [`docker-credential`](https://github.com/keirlawson/docker_credential/) library is already doing that, plus it's also handling the case of tokens being used, instead of username/passwords.

The `docker-credential` has these good points:
  * small size
  * super small dependencies (just serde and base64)
  * well written
  * has tests too (I like the approach the author took to [test the external helpers](https://github.com/keirlawson/docker_credential/blob/master/src/lib.rs)

I'm not concerned by the lack of activity over the repository. I think this is the clear example of a project that has a really specific goal and is dealing with something that is stable. The format of the docker config file hasn't been changed since years, the same applies to how external credential helpers work. The way I see it, this project does it job in a good way, hence there's no need to touch it.

I already used this library for `krew-wasm`, it's easy to use. I wonder how hard it would be to make it fit into our source code. We have public types exposed to the consumers of `policy-fetcher`, and then we have internal types used to access the docker information. I think that, thanks to this approach, we could keep using the public types (hence no need to change other libraries we have), and just swap the internal types with the ones provided by this crate.

src/registry/config.rs
viccuad: Thanks Raul! Indeed, with the ["Credential Helper Protocol"](https://docs.docker.com/engine/reference/commandline/login/) listed there it should be ok, and that covers my concerns about the `get`.

With that I'm happy to punt on this topic.

Still, we are only supporting creds store and not credential helpers, so I would either change the function names and docs and merge as it is, or as you have said, implement support for credential helpers . If the second, I favour the crate, as it is quite minimal.

viccuad: on the topic of credential helpers and store.. AFAIU  their binaries will be present when using `kwctl`, but not in the `policy-server` image? Users will then ask themselves why does it work with one and not the other.

So I suppose we should add all the helpers and stores to the policy-server image?

raulcabello: ok, I'll try to use this crate

raulcabello: > on the topic of credential helpers and store.. AFAIU their binaries will be present when using kwctl, but not in the policy-server image? Users will then ask themselves why does it work with one and not the other.

> So I suppose we should add all the helpers and stores to the policy-server image?

yes we can add them to the policy-server image, however I'm not sure if that would be enough as the container also needs to have the login credentials for the store. For example, if you are using aws you need to be able to login to ecr in addition to have the `docker-credentials-ecr-login` binary. I'm not sure how it works if you are using IAM role, maybe it is using an identity token for login? which we don't support at the moment. We need to look into this.

viccuad: oh, true. I'm happy adding support for `kwct`l only for now, maybe the users can provide a specific one time `docker/config.json` for the policy-server, that doesn't depend on credential helpers. We can't be the only ones hitting this.

flavio: Yes, I would start by worrying about the `kwctl` use case. We can think about policy-server later on, based on user feedback too.

raulcabello: We are creating a `DockerConfig` at the beginning of `policy-server` and `kwctl` that is always used and never recreated. This is not supported in the crate as it reads the config each time, [this](https://github.com/keirlawson/docker_credential/blob/master/src/lib.rs#L100) is the only pub method. In my opinion it makes sense to read it each time just in case there was a change in the config. 

For the `.docker/config.json` the crate will look for the `$DOCKER_CONFIG` env var or in the home dir if not present, and there is no way to pass it as a String. We are currently using the `docker-config-json-path` arg, I would change this to use the `DOCKER_CONFIG` env var, which means users would have to set the env var before using kwctl. 
For `policy-server` we would need to change [this env var](https://github.com/kubewarden/kubewarden-controller/blob/main/internal/pkg/admission/policy-server-deployment.go#L312) to `DOCKER_CONFIG`. This should work fine as long as the new version of the `controller` is used with the new version of `policy-server`. 

In policy-fetcher it would look like [this](https://github.com/raulcabello/policy-fetcher/commit/2a5f6a27c792374d123d8eca7820063f7418944b) . We would have to make changes in `policy-evaluator`, `policy-server` and `kwctl` as our current `DockerConfig` is used in many places, that's why I wanted to check with you before making changes in all projects.

What do you think?

flavio: > We are creating a DockerConfig at the beginning of policy-server and kwctl that is always used and never recreated. This is not supported in the crate as it reads the config each time, [this](https://github.com/keirlawson/docker_credential/blob/master/src/lib.rs#L100) is the only pub method. In my opinion it makes sense to read it each time just in case there was a change in the config.

I think doing that every time would make sense, because the basic auth credentials provided by the AWS helper change over the time. However, keep in mind that for our use case we can just read the credentials once:
  * kwctl: it pull/push/run a single policy at a time. Being a short lived process, the credentials are read every time
  * policy-server: the pollicies are pulled only once, during the bootstrap phase. Hence I think it's safe to read the credentials once and then have the downloader use them to download all the policies

> For the .docker/config.json the crate will look for the $DOCKER_CONFIG env var or in the home dir if not present, and there is no way to pass it as a String. We are currently using the docker-config-json-path arg, I would change this to use the DOCKER_CONFIG env var, which means users would have to set the env var before using kwctl.
For policy-server we would need to change [this env var](https://github.com/kubewarden/kubewarden-controller/blob/main/internal/pkg/admission/policy-server-deployment.go#L312) to DOCKER_CONFIG. This should work fine as long as the new version of the controller is used with the new version of policy-server.

Another option could be to keep all our flags and environment variables, and then have the rust code change the `DOCKER_CONFIG` env variable accordingly. I would have our ENV variable and flag have the precedence over the value specified inside of the `DOCKER_CONFIG`.

I've done a simple test:

```rust
fn main() {
    const FOO_VAR: &str = "FOO";

    println!("before: {:?}", std::env::var(FOO_VAR).ok());

    std::env::set_var(FOO_VAR, "hello");

    println!("after: {:?}", std::env::var(FOO_VAR).ok());
}
```

As expected, the process changes its own `FOO` variable, without affecting the shell environment:

```console
╭─ flavio@ganymede  ~/env-test   master ?                                                                                                                                                                                  ✔  11:55:02 
╰─ export FOO=ciao        
╭─ flavio@ganymede  ~/env-test   master ?                                                                                                                                                                                  ✔  11:55:13 
╰─ ./target/debug/env-test
before: Some("ciao")
after: Some("hello")
╭─ flavio@ganymede  ~/env-test   master ?                                                                                                                                                                                  ✔  11:55:16 
╰─ echo $FOO              
ciao
╭─ flavio@ganymede  ~/env-test   master ?                                                                                                                                                                                  ✔  11:55:18 
╰─ unset FOO
╭─ flavio@ganymede  ~/env-test   master ?                                                                                                                                                                                  ✔  11:55:23 
╰─ echo $FOO

╭─ flavio@ganymede  ~/env-test   master ?                                                                                                                                                                                  ✔  11:55:24 
╰─ ./target/debug/env-test
before: None
after: Some("hello")
╭─ flavio@ganymede  ~/env-test   master ?                                                                                                                                                                                  ✔  11:55:27 
╰─ echo $FOO              

╭─ flavio@ganymede  ~/env-test   master ? 

```

raulcabello: > Another option could be to keep all our flags and environment variables, and then have the rust code change the DOCKER_CONFIG env variable accordingly. I would have our ENV variable and flag have the precedence over the value specified inside of the DOCKER_CONFIG.

I like that approach

raulcabello: please have a look into the latest commit that replaces `DockerConfig` with https://github.com/keirlawson/docker_credential

tests/test_data/creds-store-present.json
```diff
@@ -0,0 +1,11 @@
...
+  "credsStore" : "store",
+  "auths": {
+    "https://index.docker.io/v1/": {
+      "auth": "dXNlcm5hbWU6dG9rZW4="
+    }
+  },
+  "HttpHeaders": {
+    "User-Agent": "Docker-Client/19.03.15 (linux)"
+  }
+}
```
flavio, line 11: I think we can remove this file, this is no longer used. Is it?

Cargo.toml
```diff
@@ -32,8 +32,11 @@ tracing = "0.1.36"
...
 walkdir = "2"
 rayon = "1.5.3"
+docker_credential = "1.0.1"
 
 [dev-dependencies]
 rstest = "0.15.0"
 tempfile = "3.2.0"
 textwrap = "0.15.0"
+mockall = "0.11.2"
+mockall_double = "0.3.0"
```
flavio, line 42: These can be removed, isn't it?

flavio (commented):
Great improvement, it always feels great to drop code!

Please, go through the comments, then I think we can merge it.

src/registry/mod.rs
```diff
@@ -88,29 +85,33 @@ impl From<ClientProtocol> for ClientConfig {
...
+                }
+                DockerCredential::UsernamePassword(user_name, password) => {
+                    RegistryAuth::Basic(user_name, password)
+                }
+            },
+            Err(error) => {
+                warn!(
+                    "Error: couldn't fetch credentials: {} . Using anonymous instead",
+                    error
+                );
```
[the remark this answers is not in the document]
raulcabello: done

tests/test_data/creds-store-present.json
raulcabello: done

Cargo.toml
raulcabello: done

PR #127, opened 2022-11-24
refactor: expose reusable code

flavio:
Policy evaluator implements all the sigstore verification code by leveraging the Verifier defined inside of this crate.

That's because the types of verification constraints exposed to our policy authors match exactly with what the verifier is capable of.

However, now we need to offer to our policy authors the ability to verify a signature produced by a certificate. This is something that is not relevant for the Verifier (which sole purpose is to verify the policies being pulled by kwcl and policy-server).

Because of that, some internal code of the verifier is now exposed as a reusable public function.

A new constructor has also been created, which allows to reuse a Cosign client. This is needed by policy-evaluator.

This is part of the work needed to address https://github.com/kubewarden/verify-image-signatures/issues/39

**Note:** I'm currently building this PR against the latest commit of sigstor-rs. I'm going to tag a new release of the sigstore crate in the next days (maybe even today)

### DIFF
--- a/src/verify/mod.rs
+++ b/src/verify/mod.rs
@@ -12,12 +12,14 @@ use oci_distribution::Reference;
 use sigstore::errors::SigstoreError;
 use std::convert::TryFrom;
 use std::str::FromStr;
+use std::sync::Arc;
+use tokio::sync::Mutex;
 use tracing::{debug, error, info, warn};
 
 /// This structure simplifies the process of policy verification
 /// using Sigstore
 pub struct Verifier {
-    cosign_client: sigstore::cosign::Client,
+    cosign_client: Arc<Mutex<sigstore::cosign::Client>>,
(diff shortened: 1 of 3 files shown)


src/verify/mod.rs
```diff
@@ -55,6 +57,18 @@ pub enum FulcioAndRekorData {
 }
 
 impl Verifier {
+    /// Craetes a new verifier that leverages an already existing
```
viccuad, line 60:
```suggestion
    /// Creates a new verifier that leverages an already existing
```

viccuad (approved):
LGTM!

flavio:
Moved to external blocked: I'm trying to tag a new release of sigstore-rs so that we don't build against the `main` branch

flavio:
rebuilding against the latest release of sigstore

Issue #131, opened 2022-12-12
Registry scheme is parsed from domain


kravciak:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When I pull policy from url in a form `domain:port/...` then domain is parsed as scheme.

```bash
~ kwctl pull ghcr.io/kubewarden/policies/psp-capabilities:latest
OK
~ kwctl pull ghcr.io:443/kubewarden/policies/psp-capabilities:latest
Error: unknown scheme: ghcr.io
```

It makes sense to add `registry://` to the kwctl pull command, but it's not always possible.
For exemple I have a problem when I want to set up airgap from local registry for `policyServer.insecureSources` parameter. I can't use `registry://` prefix there and if I use port then domain is parsed as scheme.
```bash
registry=registry-bge:5000
helm install kubewarden-defaults ./kubewarden-defaults --set common.cattle.systemDefaultRegistry=$registry \
  --set recommendedPolicies.enabled=True \
  --set policyServer.insecureSources={$registry}

kubectl  logs -n kubewarden policy-server-default-7bd89b4c87-b2grw
2022-12-12T12:43:54.506439Z  INFO policy_server::policy_downloader: policies download download_dir="/tmp" policies_count=6 status="init"
2022-12-12T12:43:54.506543Z ERROR policy_server: error while downloading policy clusterwide-no-host-namespace-sharing from registry-bge:5000/kubewarden/policies/host-namespaces-psp:v0.1.2: unknown scheme: registry-bge
```

PR #137, opened 2023-01-13
fix: correct parsing of some policy urls

flavio:
Prior to this commit, a policy url like `ghcr.io:443/kubewarde/policy1:latest` would cause the url scheme to be `ghcr.io`.

This commit urls with port numbers are always handled correctly.

Fixes https://github.com/kubewarden/policy-fetcher/issues/131


jvanz (approved):
LGTM! Thanks for the comprehensive tests.

viccuad (approved):
LGTM!

Issue #149, opened 2023-08-21
Fix tests failing on x86_64-pc-windows-*


fabriziosestito:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When running tests on windows, we have some failures:

```
failures:
    store::tests::keep_policy_full_path_unix
    store::tests::policy_full_path::case_1
    store::tests::policy_full_path::case_2
    store::tests::policy_full_path::case_3
    store::tests::policy_full_path::case_4
    store::tests::policy_path::case_1
    store::tests::policy_path::case_2
    store::tests::policy_path::case_3
    store::tests::policy_path::case_4
    tests::store_pull_destination_from_http
    tests::store_pull_destination_from_http_with_port
    tests::store_pull_destination_from_https
    tests::store_pull_destination_from_https_with_port
    tests::store_pull_destination_from_registry
    tests::store_pull_destination_from_registry_with_port
 ```
 
 Most of the failures are related to the platform dependant code in: https://github.com/kubewarden/policy-fetcher/blob/main/src/store.rs#L159 
 
 
 Related to #133 

### Expected Behavior

Tests should pass.

We might want to add specific tests for windows.

### Steps To Reproduce

With cargo and wine run:

`cargo test --target x86_64-pc-windows-gnu`

Using `cross-rs`, run:

`cross test --target x86_64-pc-windows-gnu`

Using `cargo-xwin` run:

`cargo xwin test --target x86_64-pc-windows-msvc`

### Environment

```markdown
- OS:
Windows
- Architecture:
x86_64
```

Issue #133, opened 2022-12-14
Feature Request: ran unit tests against windows and Mac


flavio:
### Is your feature request related to a problem?

It would be great to have the CI run the PR unit tests also on the windows and Mac platforms.

The windows platform is especially important since the library has some platform dependent code

PR #150, opened 2023-08-28
refactor: refactor platform dependant code and fix windows tests

fabriziosestito:
## Description

This PR refactors platform dependant code on the `Store` struct, by separating it in a different module and using `cfg_if` to switch the implementation when the target is Windows. (see a similar approach used here: https://github.com/google/crosvm/blob/main/src/sys.rs).
This allows to clean-up the `Store` struct from platform-dependent code and separate unit tests in a specific module.

The refactored functions use generics (`AsRef<Path>` trait implementers) so they can be used in various methods of the `Store` struct and in the tests. 

This fixes failing tests on Windows closing: #149 and adds the CI targets `windows-latest` and `macos-latest` to the testing step, closing: #133

## Test

Test on Windows should pass.

## Additional Information

Windows filesystems are notably picky about the filenames, so we encode path components to base64.
NOTE: It seems we don't encode the "scheme" part of the path, even if the [old unit tests are testing against this](https://github.com/kubewarden/policy-fetcher/blob/1f64efdd36323aee7ffdeb1100c406a938a74767/src/store.rs#L337), so I needed to add some logic to keep backward compatibility and spit known scheme-related code to a separate module.

### Potential improvement

Probably a better option to consider for the future could be to use file/directory name sanitization via [sanitize_filename](https://docs.rs/sanitize-filename/latest/sanitize_filename/).
This way we could get rid of platform-dependent code.
This breaks existing stores so we should consider it carefully.

I will send another PR adding integration tests for Store methods such as `list` that are currently missing.


.github/workflows/tests.yml
```diff
@@ -19,7 +19,10 @@ jobs:
 
   test:
     name: Test Suite
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
```
jvanz, line 25: :+1:

jvanz (approved):
LGTM

flavio (approved):
Thanks for these changes. The code is much better organized and tested now

:clap:

flavio:
I think this PR goes in the right direction, but I started thinking about the proposal to use `sanitize_filename`. I think this would be a bad idea, let me explain why.

Currently this crate is used by policy-server and by kwctl.

kwctl doesn't have a database that keeps track of the policies that have been downloaded. When `kwctl policies` is invoked we basically list all the contents of the store and build our knowledge of the policies that are available. By using the `sanitize_filename` crate we could run into cases where the original name of the policy is lost forever.

Talking about policy-server, I think it would be possible for an attacker to come up with a policy name that, once sanitized, clashes with another policy name. Looking at our Kubernetes integration, we always have a `default` Policy Server running. Admin users and regular users end up scheduling policies over there by default. The admin users can schedule both `ClusterAdmissionPolicy` and the namespaced `AdmissionPolicy`. Users with the right RBAC privileges can schedule `AdmissionPolicy` objects. We could imagine a scenario where an attacker leverages this "sanitization name clash" to overwrite a `ClusterAdmissionPolicy` wasm file with another one.

## Maybe it's time to grow up?

Docker, podman and others use a key-value store to keep track of the container images (and layers) being downloaded. We didn't want to go that way in the beginning, because we wanted to keep things simple.

However, let's imagine we would have a database with a table like that:

| policy name | shasum |
|-------------------|-------------|
| ghcr.io/kubewarden/privileged-policy:0.1.0 | foobar1 |
| ghcr.io/kubewarden/privileged-policy:0.1.1 | foobar2 |
| registry.local.lan/policies/privileged-policy:0.1.1 | foobar2 |

This would allow us to save all the policies inside of a flat directory:

```
~/.cache/kubewarden/store/
 ├──  foobar1
 └──  foobar2
```

There are a couple of advantages to this approach:
- Data deduplication: policies are not as big as container images, but this could save disk space in certain circumstances
- No need to sanitize protocol, host, path to policy and policy name (+ version). Everything would be saved using its shasum

We could use sqlite to keep the database. It's battle tested and allows multiple processes to access the sqlite file concurrently (see [here](https://www.sqlite.org/faq.html#q5)). This kind of access is relevant only to kwctl, not to policy server. To be honest, this would be an improvement over the current implementation that is not safe with regards to concurrent access to the store.
I've already used sqlite inside or a rust project and I was also able to come up with a statically linked binary (a requirement for kwctl and policy-server).

Of course this change would be quite a breaking one, but that would apply only to kwctl. Currently policy-server uses an ephemeral store when ran inside of Kubernetes, hence policies are always downloaded at startup time. For kwctl we could envision a migration step done at application startup, just once, when the sqlite database is not found.

PR #151, opened 2023-08-29
test: add store list integration test

fabriziosestito:
## Description

Adds Store integration tests, specifically it adds tests for the `list` method.

## Test

Tests should pass.

## Additional Information

I've used `tempfile` rather than fixtures in the repository since using characters as `:` in the filenames would prevent checking out the repository on Windows machines/

Depends on: #150


flavio (approved):
LGTM

PR #152, opened 2023-08-31
feat: add Policy getters to Store

fabriziosestito:
## Description

Adds `Store::get_policy_by_uri` and `Store::get_policy_by_sha_prefix`.
The latter takes a partial sha and returns a policy that matches.
Returns error if multiple results are found.
This behavior is similar to the docker cli, see: https://github.com/docker/engine/blob/8955d8da8951695a98eb7e15bead19d402c6eb27/daemon/container.go#L36

## Test

Added integration tests.

### DIFF
--- a/tests/store.rs
+++ b/tests/store.rs
@@ -6,8 +6,8 @@ use policy_fetcher::store::{path, Store};
 use tempfile::tempdir;
 
 #[test]
-fn test_list() -> Result<()> {
-    let store_root = tempdir()?;
+fn test_list() {
+    let store_root = tempdir().unwrap();
 
     let mut expected_policies = vec![
         Policy {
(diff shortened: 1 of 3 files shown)


tests/store.rs
```diff
@@ -44,6 +44,123 @@ fn test_list() -> Result<()> {
     Ok(())
 }
 
+#[test]
+fn test_get_policy_by_uri() -> Result<()> {
```
flavio, line 48: I'm fine with that, I just want to share I have always a mixed feeling about tests that return a `Result`. I don't know if it's better to avoid the usage of `?` and instead use `.expect("meaningful error about what was expected")`.

tests/store.rs
```diff
@@ -44,6 +44,123 @@ fn test_list() -> Result<()> {
...
+    assert!(result.is_none());
+
+    Ok(())
+}
+
+#[test]
+fn test_get_policy_by_uri_unknown_scheme() -> Result<()> {
+    let store_root = tempdir()?;
+
+    let policies = vec![Policy {
```
flavio, line 84: I think clippy might complain because latest version of rust allows to declare a `[ ]` without using the `vec!` macro.

tests/store.rs
```diff
@@ -44,6 +44,123 @@ fn test_list() -> Result<()> {
...
+
+    let result = store.get_policy_by_sha_prefix("93a")?;
+
+    assert!(result.is_none());
+
+    Ok(())
+}
+
+#[test]
+fn test_get_policy_by_sha_prefix_duplicate() -> Result<()> {
```
flavio, line 136: Again, I think everything is good. I just wonder if using something like the `rstest` crate could DRY the tests

flavio (approved):
LGTM, I left some comments that can be totally ignored

tests/store.rs
fabriziosestito: I agree, I just tried to use the same style that was used in the other tests. I could change this and then we can use the same approach when refactoring the rest. 
TBH I usually `unwrap()` in tests when I am sure to obtain an `Ok(T)`, as long as I don't have something specific to expect. Wdyt?

tests/store.rs
fabriziosestito: Thanks!

tests/store.rs
flavio: I like the usage of `unwrap` or `expect` over the `-> Result<>` approach

PR #153, opened 2023-09-01
fix: list non existing store

fabriziosestito:
## Description

In the previous implementation, the `read_dir` error was silenced, so when a store with a non-existing root was listed an empty list was returned. See: https://github.com/kubewarden/policy-fetcher/blob/3f73d90678939b7375f5f4cb2ad76ecdec1dfe2f/src/store.rs#L189
Since we now bubble up any error caused by `read_dir`, this was returning an error if the path did not exist breaking the original behavior.
This was breaking e2e tests on kwctl too.

## Test

Added an integration test for this scenario.

## Additional Information

NOTE: This implementation is probably going to change soon with the Database feature.

### DIFF
--- a/src/store/mod.rs
+++ b/src/store/mod.rs
@@ -123,6 +123,10 @@ impl Store {
     pub fn list(&self) -> Result<Vec<Policy>> {
         let mut policies = Vec::new();
 
+        if !self.root.exists() {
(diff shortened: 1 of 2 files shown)


flavio (approved):
LGTM

PR #154, opened 2023-09-04
chore: do not use Result in tests

fabriziosestito:
## Description

As mentioned in this PR comment, we prefer not to use `Result` in tests: 

https://github.com/kubewarden/policy-fetcher/pull/152#discussion_r1311891401

This PR refactors the existing tests that do not comply.

## Test

-

## Additional Information

Maybe we could create a coding convention and guideline docs for all the rust projects.
This doc could collect testing-related information too: e.g. helper libraries to use, test organization, testing patterns, etc...

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -391,96 +391,102 @@ mod tests {
     }
 
     #[test]
-    fn local_file_pull_destination_excluding_filename() -> Result<()> {
+    fn local_file_pull_destination_excluding_filename() {
         assert_eq!(
             pull_destination(
-                &Url::parse("https://host.example.com:1234/path/to/policy.wasm")?,
-                &PullDestination::LocalFile(std::env::current_dir()?),
-            )?,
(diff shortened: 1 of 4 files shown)


viccuad (approved):
LGTM!

About the coding convention, would be nice, but in my opinion if we have it we should enforce it and refactor things. And that's quite the work to do.

I feel that with our modular and minimal architecture, the codebase is in actually good shape. Of course we can revisit some parts and do light refactors here and there, but I don't feel we are handicapped by the current approach, unless we get an explosion of contributions that diminish the quality. So I vote to postpone such a convention.

jvanz:
> As for the coding conventions, I think it would be nice to have some guideline. That wouldn't mean we would have to go out and fix all the existing codebase to ensure everything is compliant with them. Let's discuss that with the rest of the team

Maybe we can write something in our `.github` repo and add a link to the  CONTRIBUTING guidelines.

PR #156, opened 2023-09-11
fix: add latest tag when no tag is present

fabriziosestito:
## Description

Resolves: https://github.com/kubewarden/kwctl/issues/554

Adds `:latest` to the destination filename if no tag is present.

## Test

Will test in kwctl e2e tests

## Additional Information

-


flavio (approved):
LGTM. BTW, it would  be  nice to have this fix released as part of kubewarden 1.7 release

PR #162, opened 2023-11-21
feat: custom error types.

### LINKED ISSUE #160: tech debt: implement custom error types
Currently the crate makes use of `anyhow::Result` to handle errors. We should implement crate-specific errors using `this-error`, like we already do with other projects we have.

jvanz:
## Description

Add custom error types for all errors generated in the library.

Fix #160


jvanz:
@kubewarden/kubewarden-developers THIS IS NOT READY. I would like to refine the error names and refactor the tests to follow our latest standards. ;)

jvanz:
> @kubewarden/kubewarden-developers THIS IS NOT READY. I would like to refine the error names and refactor the tests to follow our latest standards. ;)

@kubewarden/kubewarden-developers I've decided to mark this PR as ready for review now. I'll open another one fixing the tests. Just to avoid mixing two different fixes into a single PR

flavio:
> @kubewarden/kubewarden-developers I've decided to mark this PR as ready for review now. I'll open another one fixing the tests. Just to avoid mixing two different fixes in a single PR

From what I can see, only the Windows tests are broken and they are broken because of the changes introduced by this PR. Do not open a new PR, but rather fix the broken windows tests over there.
We can still review the PR in the meantime, just push the windows fixes into a dedicated commit.

src/https.rs
```diff
@@ -1,36 +1,51 @@
...
 // Struct used to reference a WASM module that is hosted on a HTTP(s) server
 #[derive(Default)]
 pub(crate) struct Https {}
 
 impl TryFrom<&Certificate> for reqwest::Certificate {
-    type Error = anyhow::Error;
+    type Error = FetcherErrors;
 
-    fn try_from(certificate: &Certificate) -> Result<Self> {
+    fn try_from(certificate: &Certificate) -> errors::Result<Self> {
```
flavio, line 24: Having defined the return type to be `errors::Result` (which implies `Result<Self, FetcherErrors>` I **think** we can void to define the type alias on line 22.

Cargo.toml
flavio, line 1: Can't we drop `anyhow` now?

src/lib.rs
```diff
@@ -16,6 +16,7 @@ pub mod sources;
 pub mod store;
 pub mod verify;
 
+use crate::errors::FetcherErrors;
```
flavio, line 19: I think we should also `use` the `Result` type define inside of the `errors` crate, just to avoid typing `errors::Result`

src/registry/mod.rs
```diff
@@ -163,12 +170,13 @@ impl Registry {
...
-        let url = Url::parse(destination).map_err(|_| anyhow!("invalid URL: {}", destination))?;
+    ) -> errors::Result<String> {
+        let url = Url::parse(destination)
+            .map_err(|_| FetcherErrors::InvalidURLError(destination.to_owned()))?;
         let sources: Sources = sources.cloned().unwrap_or_default();
-        let destination = destination
-            .strip_prefix("registry://")
-            .ok_or_else(|| anyhow!("Invalid destination format"))?;
+        let destination = destination.strip_prefix("registry://").ok_or_else(|| {
+            FetcherErrors::InvalidDestinationError("Invalid destination format".to_owned())
```
flavio, line 178: The error message seems pretty specific and almost a translation of the error type. If this error message is not going to change across the invocations (which seems to be the case), just move the string into the error type definition.

src/registry/mod.rs
```diff
@@ -151,7 +158,7 @@ impl Registry {
         Registry::client(cp)
             .fetch_manifest_digest(&reference, &registry_auth)
             .await
-            .map_err(|e| anyhow!(e))
+            .map_err(FetcherErrors::OCIRegistryError)
```
flavio, line 161: I think, given the `FetcherErrors::OCIRegistryError` builds from the OCI error, this line can be removed and replaced with a simple `?`

src/errors.rs
```diff
@@ -0,0 +1,69 @@
...
+#[derive(Error, Debug)]
+pub enum FetcherErrors {
+    #[error("{msg}: {source}")]
+    LoadCertificateError {
+        msg: String,
+        #[source]
+        source: reqwest::Error,
+    },
+    #[error("certificate {0} is not in PEM nor in DER encoding")]
+    LoadRawCertificateError(String),
```
flavio, line 14: Maybe we can conflate the two errors into a single one: `InvalidCertificateError` or something like that

src/registry/mod.rs
```diff
@@ -237,24 +242,18 @@ impl Registry {
             )
             .await
             .map(|push_response| push_response.manifest_url)
-            .map_err(|e| anyhow!("could not push policy: {}", e))
+            .map_err(FetcherErrors::OCIRegistryError)
```
flavio, line 245: I think it could be replaced with a `?`

src/registry/mod.rs
```diff
@@ -237,24 +242,18 @@ impl Registry {
...
+pub(crate) fn build_fully_resolved_reference(url: &str) -> errors::Result<Reference> {
     let image = url.strip_prefix("registry://").unwrap_or(url);
-    Reference::try_from(image).map_err(|e| {
-        anyhow!(
-            "Cannot parse {} into an OCI Reference object: {:?}",
-            image,
-            e
-        )
-    })
+    Reference::try_from(image).map_err(FetcherErrors::InvalidOCIImageReferenceError)
```
flavio, line 251: Same here, maybe a `?` is enough

src/sources.rs
```diff
@@ -101,18 +99,15 @@ impl TryFrom<RawSources> for Sources {
...
         if reqwest::Certificate::from_pem(raw_certificate.0.as_bytes()).is_ok() {
             Ok(Certificate::Pem(raw_certificate.0.as_bytes().to_vec()))
         } else if reqwest::Certificate::from_der(raw_certificate.0.as_bytes()).is_ok() {
             Ok(Certificate::Der(raw_certificate.0.as_bytes().to_vec()))
         } else {
-            Err(anyhow!(
-                "certificate {:?} is not in PEM nor in DER encoding",
-                raw_certificate
-            ))
+            Err(FetcherErrors::LoadRawCertificateError(raw_certificate.0))
```
flavio, line 110: This was already there before, but I think it's not going to be that useful to dump the whole data into the error. This could be something massive (in terms of size).

Keep that in mind if you conflate the `LoadCertificateError` and the `LoadRawCertificateError` together

src/store/mod.rs
```diff
@@ -1,12 +1,14 @@
...
 use directories::ProjectDirs;
 use lazy_static::lazy_static;
 use path_slash::PathExt;
 use std::path::{Path, PathBuf};
 use url::Url;
 use walkdir::WalkDir;
 
-use crate::policy::Policy;
+use crate::{
+    errors::{self, FetcherErrors},
```
flavio, line 9: yep, the more I look into that the more I think we have to use also `Result` to make the final code look nicer

src/store/path/windows.rs
```diff
@@ -42,7 +42,7 @@ pub fn encode_filename(filename: &str) -> String {
 }
 
 /// Decode a path that was transformed with `encode_path`.
-pub fn decode_path<P: AsRef<Path>>(path: P) -> Result<PathBuf> {
+pub fn decode_path<P: AsRef<Path>>(path: P) -> errors::Result<PathBuf> {
```
flavio, line 45: According to the GH action, line 61 below is the one causing the windows tests to fail. Seems an easy fix :)

We are not handling the `BASE64_ENGINE.decode` error

src/verify/config.rs
```diff
@@ -183,14 +185,15 @@ pub fn build_latest_verification_config(config_str: &str) -> Result<LatestVerifi
...
+                .map_err(FetcherErrors::InvalidVerifyFileError);
+            return err;
         }
     };
 
     if config.all_of.is_none() && config.any_of.is_none() {
-        return Err(anyhow!(
-            "config is missing signatures in both allOf and anyOff list"
+        return Err(FetcherErrors::InvalidVerifyFileConfigurationError(
+            "config is missing signatures in both allOf and anyOff list".to_owned(),
```
flavio, line 196: I think these two error messages can be conflated into a single one, which takes a `String` as parameter. In the 1st case we can return the serde error, in the second one we put the the string we have written.

src/verify/mod.rs
```diff
@@ -166,22 +166,18 @@ impl Verifier {
...
             None => policy.uri.as_str(),
             Some(url) => url,
         };
         if let Err(e) = Reference::try_from(image_name) {
-            return Err(anyhow!(
-                "Verification only works with OCI images: Not a valid oci image {}",
-                e
-            ));
+            return Err(FetcherErrors::InvalidOCIImageReferenceError(e));
         }
```
flavio, line 176: Not on you because it was already there, but I think we can write this in a more idiomatic way:

```rust
Reference::try_from(image_name).map_err(FetcherErrors::InvalidOCIImageReferenceError)?;
```

src/verify/mod.rs
```diff
@@ -332,10 +328,7 @@ pub async fn fetch_sigstore_remote_data(
         Some(url) => url,
     };
     if let Err(e) = Reference::try_from(image_name) {
-        return Err(anyhow!(
-            "Verification only works with OCI images: Not a valid oci image {}",
-            e
-        ));
+        return Err(FetcherErrors::InvalidOCIImageReferenceError(e));
```
flavio, line 331: Same here, I think we can rewrite that with `map_err()?`

src/errors.rs
flavio, line 1: I wonder if using the `source` macro of `thiserror` causes our API to "leak" the error types of the other crates. Is that a good thing? Should we do that? I guess there's no problem with errors that are part of the stdlib, but what about external crates?

flavio (changes requested):
Overall LGTM, I left some comments around

src/lib.rs
jvanz: But then we will start to do what we are trying to avoid while using anyhow. From our code conventions:
```
Qualify anyhow::Result rather than use anyhow::Result.
```

fabriziosestito (changes requested):
Thanks for taking care of this! 
I've noticed that flatting errors of different modules in the same enum could introduce a problem.
For instance, if the `Policy::digest` returns an error, the `?` will convert the `std::io::Error` in the first available `#[from] std::io::Error` in the FetcherErrors enum.
In this case `digest` will return a `CannotReadCertificateError` variant.

We could instead have separate error enums or structs  in each module, depending on the module has to deal with one or more than one error, and then use those "inner" errors inside the `FetcherErrors` enum with the transparent macro, such as:

```rust
pub enum FetcherError {
   #[error(transparent)]
   PolicyError(#[from] crate::policy:DigestError)
  #[error(transparent)]
   CertifcateError(#[from] crate::CertficateError)
   ....
}
```

In the example above we'll need to introduce the `DigestError` directly in the `policy` module and change the code as follows:

```rust
#[derive(thiserror::Error, Debug)]
#[error("cannot retrieve path from uri: {err}")]
pub struct DigestError {
    #[from]
    err: std::io::Error,
}

impl Policy {
    pub fn digest(&self) -> Result<String, DigestError> {
        let d = Sha256::digest(std::fs::read(&self.local_path)?);
        Ok(format!("{:x}", d))
    }
}
```

where `Result` is the plain `core::Result`.

This way the error will be converted from the caller in the appropriate parent error, `FetcherError` in our case, and we avoid the error conversion clashing together. Also, we improve readability.

You can check this approach in the wild in this crate:
https://github.com/oxidecomputer/omicron/blob/3f702ef442a2cb6522684c8b4028bc8a8b11ed6d/illumos-utils/src/zfs.rs#L37

Also, I see that we could improve unit testing, by adding assertions on the type of errors in the errors path.

src/https.rs
jvanz: Nops, it's not possible. I've tried that while I was changing this file. This is the error:

```console
cargo fmt --all && make test
cargo fmt --all -- --check
cargo clippy -- -D warnings
    Checking policy-fetcher v0.8.1 (/home/jvanz/suse/policy-fetcher)
error[E0046]: not all trait items implemented, missing: `Error`
  --> src/https.rs:21:1
   |
21 | impl TryFrom<&Certificate> for reqwest::Certificate {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `Error` in implementation
   |
   = help: implement the missing item: `type Error = /* Type */;`

For more information about this error, try `rustc --explain E0046`.
error: could not compile `policy-fetcher` (lib) due to previous error

```

src/lib.rs
flavio: Not if we entirely remove anyhow as a dependency of our crate. I would aim for that

flavio:
@fabriziosestito I didn't know about this approach. That looks way better than the one we use inside of other crates we contribute to!

Sounds like have to do some refactoring elsewhere :)

src/errors.rs
```diff
@@ -0,0 +1,39 @@
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum FetcherErrors {
```
fabriziosestito, line 4: nitpick: this is usually named a singular form.

src/https.rs
```diff
@@ -10,27 +9,42 @@ use url::Url;
...
 // Struct used to reference a WASM module that is hosted on a HTTP(s) server
 #[derive(Default)]
 pub(crate) struct Https {}
 
 impl TryFrom<&Certificate> for reqwest::Certificate {
-    type Error = anyhow::Error;
+    type Error = SourceErrors;
 
-    fn try_from(certificate: &Certificate) -> Result<Self> {
+    fn try_from(certificate: &Certificate) -> std::result::Result<Self, SourceErrors> {
```
fabriziosestito, line 21: we should alias this with a use  (also in the other part of the code)

src/registry/errors.rs
```diff
@@ -0,0 +1,21 @@
+use thiserror::Error;
+
+use crate::errors::InvalidURLError;
+
+#[derive(Error, Debug)]
+pub enum RegistryErrors {
```
fabriziosestito, line 6: nit: singular form

src/sources.rs
```diff
@@ -1,12 +1,32 @@
...
+use thiserror::Error;
 
+use crate::errors::FailedToParseYamlDataError;
 use std::collections::{HashMap, HashSet};
 use std::convert::{TryFrom, TryInto};
 use std::path::{Path, PathBuf};
 use std::{fs, fs::File};
 
+#[derive(Error, Debug)]
+pub enum SourceErrors {
```
fabriziosestito, line 11: nit: singular form

src/store/errors.rs
```diff
@@ -0,0 +1,19 @@
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum StoreErrors {
```
fabriziosestito, line 4: nit: singular form

src/verify/errors.rs
```diff
@@ -0,0 +1,34 @@
+use thiserror::Error;
+
+use crate::{errors::FailedToParseYamlDataError, registry::errors::RegistryErrors};
+
+#[derive(Error, Debug)]
+pub enum VerifyErrors {
```
fabriziosestito, line 6: nit: singular form

fabriziosestito (changes requested):
overall LGTM, some minor comments added.

Also, I've noticed we have some tests that check on errors.
Some example here but there could be more:

https://github.com/kubewarden/policy-fetcher/blob/30e22b0061aff2264421622e4afe4d6490b841f1/src/sources.rs#L303

https://github.com/kubewarden/policy-fetcher/blob/30e22b0061aff2264421622e4afe4d6490b841f1/src/sources.rs#L286
Could we update the test with an assertion on the type of the error?, such as
```rust 
assert!(matches!(result, Err(NewError(_)))
```

src/errors.rs
```diff
@@ -0,0 +1,41 @@
...
+    StoreError(#[from] crate::store::errors::StoreError),
+    #[error(transparent)]
+    InvalidURLError(#[from] InvalidURLError),
+    #[error(transparent)]
+    CannotCreateStoragePathError(#[from] CannotCreateStoragePathError),
+}
+
+#[derive(thiserror::Error, Debug)]
+#[error("{0}")]
+pub struct FailedToParseYamlDataError(#[from] pub serde_yaml::Error);
```
flavio, line 33: This is not "wrapped" by the `FetcherError` enum, isn't it?

src/errors.rs
```diff
@@ -0,0 +1,41 @@
...
+
+#[derive(thiserror::Error, Debug)]
+#[error("invalid URL: {0}")]
+pub struct CannotCreateStoragePathError(#[from] pub std::io::Error);
+
+#[derive(thiserror::Error, Debug)]
+#[error("invalid URL: {0}")]
+pub struct InvalidURLError(pub String);
```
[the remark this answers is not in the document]
jvanz: This error exists because it is used in multiple modules. Thus, it's not necessary to duplicate it all over the place.  And there is a wrapper around it in the `FetcherError` because the `FetcherError` is the type used in the Result returned by the lib to the callers.

src/errors.rs
jvanz: It's not. But this error exists because it used in multiple modules. Those modules errors (`SourceError` and `VerifyError`) wraps this error.

src/sources.rs
```diff
@@ -322,8 +340,13 @@ Wm7DCfrPNGVwFWUQOmsPue9rZBgO
         );
         let raw_source_authorities: RawSourceAuthorities = serde_json::from_value(raw).unwrap();
 
-        let actual: Result<SourceAuthorities> = raw_source_authorities.try_into();
-        assert!(actual.is_ok(), "Got an expected error: {:?}", actual);
+        let actual: SourceResult<SourceAuthorities> = raw_source_authorities.try_into();
+
+        assert!(
+            matches!(actual, Ok(_)),
```
fabriziosestito, line 346: `actual.is_ok()`

src/sources.rs
```diff
@@ -299,15 +324,8 @@ Wm7DCfrPNGVwFWUQOmsPue9rZBgO
...
-        match expected {
-            Ok(RawCertificate(data)) => {
-                assert_eq!(&data, expected_contents);
-            }
-            unexpected => {
-                panic!("Didn't get what I was expecting: {:?}", unexpected);
-            }
-        }
+        let expected: SourceResult<RawCertificate> = auth.try_into();
+        assert!(matches!(expected, Ok(RawCertificate(data)) if &data == expected_contents));
```
fabriziosestito, line 328: the linter in my editor complains about the needless reference

src/store/errors.rs
```diff
@@ -0,0 +1,23 @@
...
+    #[error("cannot read policy in local storage: {0}")]
+    FailedToReadPolicyInLocalStorageError(#[from] walkdir::Error),
+    #[error(transparent)]
+    PolicyStoragePathError(#[from] std::path::StripPrefixError),
+    #[error("unknown scheme: {0}")]
+    UnknownSchemeError(String),
+    #[error("multiple policies found with the same prefix: {0}")]
+    MultiplePoliciesFoundError(String),
+    #[error(transparent)]
+    DigestErrors(#[from] crate::policy::DigestError),
```
fabriziosestito, line 20: singular form

src/policy.rs
```diff
@@ -9,8 +9,15 @@ pub struct Policy {
...
+#[derive(thiserror::Error, Debug)]
+#[error("cannot retrieve path from uri: {err}")]
+pub struct DigestError {
+    #[from]
+    err: std::io::Error,
+}
+
 impl Policy {
-    pub fn digest(&self) -> Result<String, std::io::Error> {
+    pub fn digest(&self) -> std::result::Result<String, DigestError> {
```
fabriziosestito, line 20: missing alias std::result

src/errors.rs
flavio: Shouldn't we drop the `pub` attribute, or turn that into a `pub(crate)` if needed?

src/sources.rs
fabriziosestito: i am using clippy as well. probably it doesn't show because it's not running in `test` mode.

fabriziosestito: try invoking `cargo clippy --tests`

tests/store.rs
```diff
@@ -160,7 +159,7 @@ fn test_get_policy_by_sha_prefix_duplicate() {
     assert!(result.is_err());
 }
 
-fn setup_store(policies: &[Policy]) -> Result<()> {
+fn setup_store(policies: &[Policy]) -> std::result::Result<(), std::io::Error> {
```
fabriziosestito, line 162: alias with a use

jvanz:
@flavio we've decide to merge this because during the last daily call you did not seems against to it after discussing your last comment. Please, let us know if you want to change something. I'll be glad to do it.

PR #167, opened 2024-02-16
ci: Add codecov support

viccuad:
## Description

Fix https://github.com/kubewarden/kubewarden-controller/issues/641

## Test

CI, codecov message.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -66,3 +66,39 @@ jobs:
         with:
           command: clippy
           args: -- -D warnings
+
+  coverage:
+    name: coverage
+    runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


viccuad:
missing store test run, even if using the same `--workspace` flag.

flavio (changes requested):
LGTM, but you have to remove all the references to integration tests since we don't have them

viccuad:
> LGTM, but you have to remove all the references to integration tests since we don't have them

I consider the tests under tests/{store.rs, sources.rs} as integration tests (or other vague type, but still, they aren't unit tests).

Removed the single doc test that was failing the compilation, see commit comment.

flavio (approved):

PR #182, opened 2024-04-12
test: fix build on windows

flavio:
Enable long paths on windows, otherwise git checkout of tough will fail.

Once merged, I'll tag a new release and propagate that to policy-evaluator. Otherwise kwctl will not build on windows


flavio:
The fix is [working fine](https://github.com/flavio/policy-fetcher/actions/runs/8664988891/job/23762662036)

PR #183, opened 2024-04-12
feat: test if registry name is used to get credentials.

### LINKED ISSUE #181: Add integration test to ensure we use registry name to get credentials from docker config.json file.
Recently, we found a bug in the policy-fetcher that prevents users from verifying signature of container image hosted on private registries. The issue was caused because we were using the full image URL to get the registry credentials from the docker config.json file.  We fix that in the PR https://github.com/kubewarden/policy-fetcher/pull/180. However, the PR missed the tests. Therefore, we need to add it to avoid a regression. This is the issue for that.

### Acceptance criteria
- Write an integration test to cover the fix from PR #180

jvanz:
## Description

Recently, we have merged a fix to use the registry name to get the credentials to access private registry from the docker config file. Before that, the code was using the whole container image name. Therefore, no credentials was not being found and causing issues when trying to interact with the private registries.

To avoid that issue happens again, this commit adds a new unit test to avoid a regression for this bug.

Fix #181 

## Test

```shell
make test
```

### DIFF
--- a/src/registry/mod.rs
+++ b/src/registry/mod.rs
@@ -28,6 +28,9 @@ use errors::RegistryError;
 
 use self::errors::RegistryResult;
 
+#[cfg(test)]
+use mockall::automock;
+
 pub mod errors;
 
 lazy_static! {
@@ -93,6 +96,8 @@ impl From<ClientProtocol> for ClientConfig {
     }
 }
 
+#[cfg_attr(test, automock)]
+#[cfg_attr(test, allow(dead_code))]
 impl Registry {
     pub fn new() -> Registry {
         Registry {}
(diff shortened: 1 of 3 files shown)


src/registry/mod.rs
```diff
@@ -125,10 +130,10 @@ impl Registry {
     }
 
     /// Fetch the manifest of the OCI object referenced by the given url.
-    pub async fn manifest(
+    pub async fn manifest<'a>(
```
viccuad, line 133: Adding lifetimes seems overkill here. Is there no other way? Are we using MockRegistry correctly?

(Also, TIL of [Higher Rank Trait Bounds](https://doc.rust-lang.org/nomicon/hrtb.html))

flavio:
This seems an overkill, it complicates things a lot and we're doing a lot of mocking. I think we should add an integration test that:
* starts a real container registry that uses authentication. This can be done with the `distribution` registry container. There's a simple way to have auth enabled against distribution, the `oci-distribution` test suite does that, I can point to the code
* push a test policy to the registry
* pull from the registry

This could be added to kwctl, I don't think its integration test take into account authenticated registries.

We could then take the same integration test and expand it to use also sigstore, to ensure `kwctl verify` works also with authenticated registries.

I wonder if we should cover that into kwctl's integration tests, or if we should add some simple integration tests here

jvanz:
We close this PR in favor of a future PR adding integration test to cover the same fix cover on this PR

PR #186, opened 2024-05-07
E2E test to fetch signature from authenticated registry.

jvanz:
## Description

Add a test to verify if the policy fetch is able to fetch the Sigstore data from an authenticated registry.

Fix https://github.com/kubewarden/policy-fetcher/issues/181

## Test

```shell
make test
```


tests/e2e.rs
```diff
@@ -0,0 +1,194 @@
...
+    // test should not be run with custom docker config
+    // because if this test fails, the docker config will be
+    // used by other tests and they will fail too
+    std::env::var("DOCKER_CONFIG")
+        .ok()
+        .map(|_| panic!("DOCKER_CONFIG is already set"));
+
+    // add the DOCKER_CONFIG environment variable to the client
+    // so it knows where to find the credentials
+    std::env::set_var("DOCKER_CONFIG", auth_dir.path());
```
jvanz, line 162: I do not like this. But the lib used to find the config file uses env var to find the file

viccuad: I do not like that the env vars are let set for the rest of the tests. We don't consume those env vars though so I consider it fine.

tests/e2e.rs
```diff
@@ -0,0 +1,193 @@
...
+async fn load_container_image(port: u16) -> Reference {
+    let client = Client::new(oci_distribution::client::ClientConfig {
+        protocol: oci_distribution::client::ClientProtocol::HttpsExcept(vec![format!(
+            "localhost:{}",
+            port
+        )]),
+        ..Default::default()
+    });
+
+    // pulling policy image. webassembly.azurecr.io/hello-wasm:v1
```
viccuad, line 88:
```suggestion
    // pulling policy image
```

viccuad (approved):
LGTM. I would prefer the test first on the file and then the aux functions, but it's fine as it is.

jvanz:
> Mac e2e fail on testcontainers :/.

Github Mac runner image does not have Docker [installed by default](https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md). :tired_face:

jvanz:
@kubewarden/kubewarden-developers I've tried to install docker in the MacOS to run the added test on it as well. But I've run into an issue on QEMU/Colima starting. Which was consuming some time. Thus, I've decided to disable the test on Mac and leave this issue for later.

tests/e2e.rs
```diff
@@ -0,0 +1,198 @@
...
+use sigstore::cosign::Client as CosignClient;
+use sigstore::cosign::ClientBuilder;
+use sigstore::cosign::Constraint;
+use sigstore::cosign::CosignCapabilities;
+use sigstore::cosign::SignatureLayer;
+use sigstore::crypto::SigningScheme;
+use sigstore::registry::Auth;
+use sigstore::registry::ClientConfig;
+use sigstore::registry::ClientProtocol;
+use sigstore::registry::OciReference;
```
flavio, line 17: can you group all the imports together? LSP has a handy shortcut to do that (at least with vim)

tests/e2e.rs
```diff
@@ -0,0 +1,198 @@
...
+const REGISTRY_CREDENTIALS_BCRYPT: &str =
+    "user:$2y$05$WRQnMYgFDnzA/wViFUOw6uNgJpjaYemWXRD2pQTrBgfU4abxv1KdO";
+const POLICY_IMAGE_TAG: &str = "ghcr.io/kubewarden/tests/pod-privileged:v0.2.5";
+
+/// Signs an image with the given client
+async fn sign_image<'a>(
+    mut client: CosignClient<'a>,
+    image: &OciReference,
+    auth: &Auth,
+) -> Result<(), String> {
```
flavio, line 41: I would just use anyhow here, make it a dev dependency. It would simplify the code below (no need to do `map_err` for example), otherwise... instead of mapping the error and bubbling it up, just do an `expect`

tests/e2e.rs
```diff
@@ -0,0 +1,198 @@
...
+    let (cosign_signature_image, source_image_digest) = client
+        .triangulate(image, auth)
+        .await
+        .map_err(|e| e.to_string())?;
+    let mut signature_layer =
+        SignatureLayer::new_unsigned(image, &source_image_digest).map_err(|e| e.to_string())?;
+    let signer = SigningScheme::ECDSA_P256_SHA256_ASN1
+        .create_signer()
+        .expect("unexpected error");
```
flavio, line 50: should be something that provides context to the error, like "cannot create signer"

tests/e2e.rs
```diff
@@ -0,0 +1,198 @@
...
+            .with_env_var("REGISTRY_AUTH", "htpasswd")
+            .with_env_var("REGISTRY_AUTH_HTPASSWD_REALM", "Registry Realm")
+            .with_env_var("REGISTRY_AUTH_HTPASSWD_PATH", "/auth/htpasswd")
+            .with_volume(auth_dir.path().to_string_lossy().to_string(), "/auth"),
+        auth_dir,
+    )
+}
+
+/// Loads a policy image into the registry running in the given port and signs it
+async fn load_container_image(port: u16) -> Reference {
```
flavio, line 79: nitpick: `push` would be better than `load` IMHO

flavio: I would also split that into these functions:

- `pull_image(uri: &str)`
- `push_image(dest_uri: &str, image_data whatever_is_returned_by_oci_image_pull)`
- `sign_image`

tests/e2e.rs
```diff
@@ -0,0 +1,198 @@
+use base64::prelude::{Engine as _, BASE64_STANDARD_NO_PAD};
+use oci_distribution::manifest;
+use oci_distribution::secrets::RegistryAuth;
+use oci_distribution::Client;
+use oci_distribution::Reference;
+use policy_fetcher::verify::fetch_sigstore_remote_data;
```
flavio, line 6: I'm fine not running this test on macos, however the linter is complaining about a bunch of stuff. It would be great to avoid these warnings.

I see two possible solutions:
- create a `tests/mod.rs` which enables the `e2e.rs` test on non-macos arch. I wonder if that works...
- add some clippy directives to suppress warnings. these should be turned on only on mac

tests/e2e.rs
```diff
@@ -0,0 +1,198 @@
...
+// The test is disabled on macOS because the GitHub Actions runner image
+// does not have the docker installed to run the test.
+#[cfg(not(target_os = "macos"))]
+#[tokio::test]
+async fn test_fetch_sigstore_data_from_registry_with_authentication() {
+    let (registry_image, auth_dir) = setup_registry_image();
+    let docker = clients::Cli::default();
+    let container = docker.run(registry_image);
+
+    let port = container.get_host_port_ipv4(5000);
```
flavio, line 178: The `5000` should be a const

tests/e2e.rs
```diff
@@ -0,0 +1,198 @@
...
+            protocol: ClientProtocol::HttpsExcept(vec![format!("localhost:{}", port)]),
+            ..Default::default()
+        })
+        .build()
+        .expect("failed to build cosign client");
+
+    create_docker_config_file(&auth_dir, port);
+
+    let client_arc = Arc::new(Mutex::new(cosign_client));
+    fetch_sigstore_remote_data(&client_arc, &format!("registry://{}", push_image.whole()))
```
flavio, line 193: Can we test the actual pulling of the image and it's verification? Basically what kwctl pull would do.

flavio (changes requested):

tests/e2e.rs
jvanz: I think the first option is not possible because cargo compiles all the files under "tests" directory as crate running tests. 

> We create a tests directory at the top level of our project directory, next to src. Cargo knows to look for integration test files in this directory. We can then make as many test files as we want, and Cargo will compile each of the files as an individual crate.

I think if we reorg how we run all the tests here we may be able to do that. But I think that should be in a different PR. I'll try follow the second approach.

jvanz:
@kubewarden/kubewarden-developers I'm aware of the windows failing tests. I'll take a look on it next week.

flavio (approved):
LGTM. thanks for the changes

PR #194, opened 2024-06-10
feat: method to get manifest and config of an OCI image.

jvanz:
## Description

Updates the registry client adding a method to fetch the container image manifest, its digest and configuration.

Fix https://github.com/kubewarden/policy-evaluator/issues/518

## Test

To test this pull request, you can run the following commands:

```shell
make test
```


viccuad (approved):
LGTM. I would enjoy having a new unit test for `manifest_and_config`, but it's not a blocker for me.

fabriziosestito (approved):

PR #215, opened 2024-12-12
add more logging

flavio:
Add more logging statements when pushing to an insecure registry.
    
We had a user that was struggling to understand why pushing to a certain registry didn't work despite `insecure` being turned on.
    
We finally figured out that was happening because of a reverse proxy was misconfigured. However, the error obtained from the insecure proxy was silently swallowed.
 

Moreover, I fixed some clippy warnings introduced by latest release of Rust

### DIFF
--- a/src/fetcher.rs
+++ b/src/fetcher.rs
@@ -1,3 +1,5 @@
+use std::fmt;
+
 use async_trait::async_trait;
 use url::Url;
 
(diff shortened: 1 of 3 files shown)


viccuad (approved):
LGTM. Maybe we could also print the version when on debug :thinking: .

fabriziosestito (approved):

jvanz (approved):

jvanz:
@flavio Sorry, I've rebased the PR and now I'm added as committer.

PR #216, opened 2024-12-13
fix: sources cert loading

fabriziosestito:
## Description

This PR resolves issues with certificate loading from a file path. Previously, using a DER certificate 
in the sources file caused the fetcher to panic due to an attempted `from_utf8` conversion on the file data. 
The `RawCertificate` data type has been updated to `bytes`, and the serde deserialization has been fixed using `serde_bytes`.

Additionally, this PR addresses a problem with PEM/DER recognition that arose after transitioning to `rustls`. 
The issue was caused by a behavior in `reqwest`, which does not return an error with `rustls` but does so when using `native_tls_rust`: https://github.com/seanmonstar/reqwest/blob/8a2174f8a4259691b5ca0a16778427127f61373a/src/tls.rs#L141

## Test

Added an integration test.


Cargo.toml
```diff
@@ -34,6 +34,7 @@ rustls = { version = "0.23", default-features = false, features = [
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 serde_yaml = "0.9"
+serde_bytes = "0.11.15"
```
flavio, line 37: Drop the patch release, I prefer to be a bit more relaxed inside of libraries

Cargo.toml
```diff
@@ -48,6 +49,7 @@ rayon = "1.5"
 docker_credential = "1.2"
 tokio = { version = "1", default_features = false }
 cfg-if = "1.0"
+x509-parser = "0.16.0"
```
flavio, line 52: Drop the patch release, I prefer to be a bit more relaxed inside of libraries

flavio (changes requested):
LGTM, can you please fix the CI issues too?

jvanz (changes requested):
LGTM. But we need to check out the CI issues. ;)

flavio (approved):
LGTM. Thanks for the fixes!

PR #233, opened 2025-02-12
fix: http registry interactions

flavio:
Prior to this PR, some methods exposed by the `Registry` struct did not handle properly interactions with HTTP registries.

This made impossible to fetch OCI manifests and configurations from certain registries.

This commit fixes the problem and refactors the code to avoid repetitions.

New e2e tests have been added that interact with http and https registries.

Finally, some structs internals have been made public to simplify the process of creating them.

For example, the `Sources` struct could be created only by deserializing the contents of a `sources.yaml` file. This made testing (and library consumpion) really hard.
    
While working on this PR, I've also added a [taplo](https://taplo.tamasfe.dev/) configuration that can be used to format the `Cargo.toml` file in a consistent way.

### DIFF
--- a/src/registry/mod.rs
+++ b/src/registry/mod.rs
@@ -1,5 +1,9 @@
+use std::{collections::BTreeMap, convert::TryFrom, str::FromStr};
+
 use async_trait::async_trait;
 use docker_credential::DockerCredential;
+use errors::RegistryError;
+use futures::future::BoxFuture;
 use lazy_static::lazy_static;
 use oci_client::{
     client::{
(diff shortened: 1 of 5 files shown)


flavio:
ouch, I forgot to mention... everything started with me adding the possibility to add `annotations` to the OCI manifest being created by `push` operations.

This is something required to implement https://github.com/kubewarden/kwctl/issues/1044

src/registry/mod.rs
```diff
@@ -93,6 +89,45 @@ impl From<ClientProtocol> for ClientConfig {
     }
 }
 
+/// Perform the gien operation using different protocols. This is done because we want to use
```
fabriziosestito, line 92: typo

fabriziosestito (approved):

src/registry/mod.rs
flavio: fixed

flavio:
I'm going to merge and tag 0.10.0

PR #237, opened 2025-03-03
fix: change log level when pushing policy

flavio:
Use `debug` instead of `warn` level when pushing a policy. This reduces the noise we produce when doing `kwctl push`.

### DIFF
--- a/src/registry/mod.rs
+++ b/src/registry/mod.rs
@@ -262,7 +262,7 @@ impl Registry {
         annotations: Option<&BTreeMap<String, String>>,
         client_protocol: ClientProtocol,
     ) -> RegistryResult<String> {
-        warn!(client_protocol = ?client_protocol, "pushing policy");
+        debug!(client_protocol = ?client_protocol, "pushing policy");
         let reference =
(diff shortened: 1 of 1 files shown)


flavio:
This is required to get rid of the "pushing policy" message, which is really really long (it's a dump of the whole CA chain)

```console
kwctl push annotated-policy.wasm testing.registry.svc.lan/kw/rego:0.0.1
2025-03-03T10:07:20.043491Z  WARN kwctl::push: annotation is a multi-line string, it will be removed from the OCI manifest annotation="io.kubewarden.policy.usage"
2025-03-03T10:07:20.043662Z  WARN policy_fetcher::registry: pushing policy client_protocol=Https(CustomCaCertificates([Pem([... really long dump ...])]))
Policy successfully pushed: testing.registry.svc.lan/kw/rego@sha256:2394e117cb26e771793b8344821aed68f2145d74f3509e306dd3becd9ffb1834
```

fabriziosestito (approved):

viccuad (approved):

PR #257, opened 2025-05-06
chore(deps): add rust-toolchain.toml.

jvanz:
## Description

Adds the rust-toolchain.toml file to configure the build command on witch toolchain and target it should use.

Related to https://github.com/kubewarden/kubewarden-controller/issues/1093

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,4 @@
+[toolchain]
+channel    = "1.86.0"
+components = ["clippy", "rust-analyzer", "rustfmt"]
+profile    = "minimal"
--- a/updatecli/updatecli.d/update-rust-toolchain.yaml
+++ b/updatecli/updatecli.d/update-rust-toolchain.yaml
@@ -1,0 +1,45 @@
+name: Update Rust version inside of rust-toolchain file
+
+scms:
+  github:
+    kind: github
+    spec:
+      user: "{{ .github.author }}"
+      email: "{{ .github.email }}"
+      owner: "{{ requiredEnv .github.owner }}"
+      repository: "policy-fetcher"
+      token: "{{ requiredEnv .github.token }}"
+      username: "{{ requiredEnv .github.user }}"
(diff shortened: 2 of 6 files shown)


rust-toolchain.toml
```diff
@@ -0,0 +1,5 @@
+[toolchain]
+channel    = "1.86.0"
```
jvanz, line 2: @kubewarden/kubewarden-developers in the github action we always use `stable` channel. If we decided to use the same here, we can remove the updatecli script to bump the version here. What do you prefer?

rust-toolchain.toml
```diff
@@ -0,0 +1,5 @@
...
+components = ["clippy", "rust-analyzer", "rustfmt"]
+profile    = "minimal"
+targets    = ["x86_64-unknown-linux-musl"]
```
flavio, line 5: is this going to cause problems locally given we don't have a musl system?

jvanz: I don't think so. Because the github runner does not have it and the CI is happy.

jvanz: Actually, we can remove this line. As we are not using the `--target` in the cargo commands, the default target (the gnu one) is being used.

updatecli/updatecli.d/update-rust-toolchain.yaml
```diff
@@ -0,0 +1,44 @@
...
+  github:
+    kind: github
+    spec:
+      user: "{{ .github.author }}"
+      email: "{{ .github.email }}"
+      owner: "{{ requiredEnv .github.owner }}"
+      repository: "policy-fetcher"
+      token: "{{ requiredEnv .github.token }}"
+      username: "{{ requiredEnv .github.user }}"
+      branch: "{{ .github.branch }}"
```
viccuad, line 13:
```suggestion
      branch: "{{ .github.branch }}"
      commitusingapi: true # enable cryptographically signed commits
```

flavio (approved):

